Issue #46, opened 2020-06-04
Rename attributes in `VotingResult` to avoid confusion with an individual `Vote`.


davefrey:
The `VotingResult` is meant to be a _tally_ of votes for a proposal. Currently, the tally attributes are `yes` and `no`, which for me is confusing: while `VotingResult.yes` is a sum, `Vote.yes` is a boolean.

I suggest instead `votes_for` and `votes_against`: these are explicitly tallies, without interpreting the `VotingResult` context. For an individual `Vote`, `yes` remains the best attribute name.

This change should include the database columns and all mentions of the attributes. It would also require superficial rename changes in other repos:
```
liquidvotingio/api
liquidvotingio/ruby-client
liquidvotingio/deployment
liquidvotingio/liquidvotingio.github.io
```
This one is mothballed:
~~liquidvotingio/liquid-voting-admin~~

oliverbarnes:
Agreed 👍. Want to send in the PRs?

Don’t worry about admin though, it’s mothballed for now

davefrey:
Agreed to use `in_favor` and `against` as the new names.

davefrey:
The last PR for this ticket will be in liquidvotingio/liquidvotingio.github.io but I can't push, could you check if I have write on that repo? Thanks! @oliverbarnes

oliverbarnes:
Added!

davefrey:
Okay I think everything is here now. If you're good, you can merge the 4 PRs

PR #48, opened 2020-06-09
WIP Rename voting results attributes

davefrey:
Rename the `yes` and `no` fields in the voting results object to be `in_favor` and `against`.
There are changes to the db, code, and documentation.
These changes should be coordinated with coming changes in other repos (see #46).
Part of #46


davefrey:
Changing to WIP, I forgot the docker image

PR #49, opened 2020-06-10
Rename voting results attributes

davefrey:
This is all of the changes needed in this repo to rename the voting results attributes from `yes`, `no` to `in_favor`, `against`. It includes a migration for the db changes.

This PR should be coordinated with forthcoming PRs for other repos, see #46 
liquidvotingio/ruby-client
liquidvotingio/deployment
liquidvotingio/liquidvotingio.github.io

Tests pass; I haven't get done local docker testing, I need to get local docker sorted first, but otherwise I think it's good.


davefrey:
> Looking good! Thanks for submitting this 👍It's the first PR of the project :)
> 
> I'd just remove the updated mix.lock, looks like it got submitted by mistake.

Done, good catch

PR #60, opened 2020-06-26
Run formatter

### LINKED ISSUE #59: Run formatter
I notice we're not using `mix format`? It might also be a good idea to check it on CI with `mix format --check-formatted`?

niklaslong:
Part of #59.


oliverbarnes:
Looking good to me, can merge once the PR draft status is removed

niklaslong:
Ok, this is quite large, let's merge this, then #61 and I'll follow up with the refactors?

PR #66, opened 2020-07-06
api#14-add-uuid-column-to-participants-table

jinjagit:
Notes:

Participant changeset doesn't validate uuid as "required field", since we want the uuid to be generated on creation of a participant. Thus, if the uuid is deleted by subsequent changes, that would not be caught. I wonder if this means we should have 2 changesets (one for create, another for update/edit - the latter having uuid as required field).

I have not set the uuid as index, nor unique_index. Currently we have a unique_index of:
```elixir
# api/priv/repo/migrations/20200519150515_update_participant_index_on_email.exs

create unique_index(:participants, [:organization_uuid, :email],  name: :uniq_index_organization_uuid_participant_email)
```

I have also not set the uuid as primary key (currently is set to `id: "participants_pkey" PRIMARY KEY, btree (id) ...`)

### DIFF
--- a/lib/liquid_voting/voting/participant.ex
+++ b/lib/liquid_voting/voting/participant.ex
@@ -20,7 +20,7 @@ defmodule LiquidVoting.Voting.Participant do
   @doc false
   def changeset(participant, attrs) do
     required_fields = [:email, :organization_uuid]
-    all_fields = [:name, :uuid | required_fields]
+    all_fields = [:name | required_fields]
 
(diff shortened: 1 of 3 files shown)


oliverbarnes:
I'd say we don't need to include the `:uuid` in the changeset at all, since it won't change. It's immutable

jinjagit:
It certainly shouldn't be changed at any point in the api, I agree.

Issue #68, opened 2020-07-06
Make :uuid primary field for Participant


oliverbarnes:
After #67 is done

niklaslong:
Was your intention to use the UUID only externally or also internally for the assocs?

oliverbarnes:
For both, in principle. Primary id for each resource, association id, and
external reference. But am open to adding an external_uuid that could be
mutable (for frontends, for example) at some point, nothing certain yet.
Thoughts?

jinjagit:
By "primary field" are you referring to an Absinthe, GraphQL, or other specific type of field, or a more generic meaning?
(I am assuming you do not men "primary key")

niklaslong:
> Thoughts?

I think simply replacing the primary key for `id` with the `uuid` for now is the simplest. Whether we choose to use another externally will be highly dependant on how we handle the mapping of external users to our internal representation of them .

oliverbarnes:
> By "primary field" are you referring to an Absinthe, GraphQL, or other specific type of field, or a more generic meaning?
> 
> (I am assuming you do not mean "primary key")

Good question - I mean for all of the above. Primary key and id for reference internally and externally

jinjagit:
So, make uuid the primary key and primary reference field across the api. Eventually.

oliverbarnes:
Exactly. And come to think of it, this probably needs to be coordinated across the other resources (votes, delegations, results), so there are parallel PRs for each, and changes are merged and deployed around the same time and the API is kept consistent

jinjagit:
closed, as covered by [#76](https://github.com/liquidvotingio/api/pull/76)

Issue #67, opened 2020-07-06
Expose Participant :uuid field on API


oliverbarnes:
Add to graphql schema and resolvers

jinjagit:
So, we managed to get the field uuid exposed in GraphQL quite easily, in that it could be returned in a query like so:
`query { Participant(id: 2) { uuid}}`
Is that all this issue requires?
We went further and made it so that GraphQL uses uuid as it's root query for Partcipant, but I think this may be moving into the following issue (set uuid to 'primary field), and this approach still needs work on integration with the other API actions on Participant.

jinjagit:
i.e. I'm not sure we can 'add to resolvers' without setting uuid to the root GraphQL query for participant (since GraphQL will complain if we find participant by uuid, without it being the root query, though it will do it). This then breaks voting, fails test: Voting: test votes get_vote!/3, so would mean moving into working on Votes? (not just Participant)

niklaslong:
I'd do it in the following steps:

- add `uuid` field in schema  + custom type for `uuid` for all resources
- replace `id` with `uuid` as primary key + fix tests for all resources

These two could also be broken into several PRs but it might be annoying to handle the relationships if half the resources are using `id` and the other half are using `uuid`...

@oliverbarnes thoughts?

oliverbarnes:
@jinjagit I see what you mean, and @niklaslong I think your suggestion makes sense. A single PR for the former, then another for the latter

jinjagit:
closed, as moved to [#76](https://github.com/liquidvotingio/api/pull/76), following GithHub weirdness.

PR #71, opened 2020-07-08
Add uuid field in schema + custom type for uuid for all resources

jinjagit:
The Participant part of Niklas's first step for issues #67 and #68 
  "add uuid field in schema + custom type for uuid for all resources"

Note: I also changed 'delegations_received' in readme graphQL examples to camel-case (which graphQL prefers).

I thought it best to get some feedback before moving on to Vote, etc.


oliverbarnes:
Converted the PR to draft since it's still in the works and Vote etc will be added soon

README.md
```diff
@@ -119,7 +119,7 @@ Then run some [queries](https://graphql.org/learn/queries/#fields):
 query {
   participants {
     email
-    delegations_received {
+    delegationsReceived {
```
oliverbarnes, line 122: 👍

lib/liquid_voting/voting.ex
```diff
@@ -220,8 +220,48 @@ defmodule LiquidVoting.Voting do
...
+
+      iex> get_participant_by_uuid("2b041b4e-81bd-4c6f-96db-0bac9b03b991", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      %Participant{}
+
+      iex> get_participant_by_uuid("unregistered-45c1-4a65-af87-6cbbbafadb14", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      nil
+
+  """
+
+  # TODO? Remove this function, as get_participant!/2, above, has same functionality?
```
oliverbarnes, line 241: I'd say yes, but check dependencies on it.

beware the bangs, too - [Repo.get_by!](https://hexdocs.pm/ecto/Ecto.Repo.html#c:get_by!/3) raises an exception when it doesn't find a record, whereas [Repo.get_by](https://hexdocs.pm/ecto/Ecto.Repo.html#c:get_by/3) returns `nil`. So they're not exactly the same. The question of when to use exceptions might popup, and the answer is: it's complicated. In principle they should be only used when something unexpected happens, but there's grey areas and different schools of thought... on going discussion.

lib/liquid_voting/voting.ex
```diff
@@ -237,13 +277,14 @@ defmodule LiquidVoting.Voting do
...
   """
+
   def get_participant_by_email(email, organization_uuid),
     do: Repo.get_by(Participant, email: email, organization_uuid: organization_uuid)
 
   @doc """
   Gets a single participant for an organization uuid by their email
 
-  Returns nil if the Participant does not exist.
+  Raises `Ecto.NoResultsError` if the Participant does not exist.
```
oliverbarnes, line 287: 👍

lib/liquid_voting/voting.ex
```diff
@@ -320,7 +362,9 @@ defmodule LiquidVoting.Voting do
       {:error, %Ecto.Changeset{}}
 
   """
-  def delete_participant(%Participant{} = participant), do: Repo.delete(participant)
+  def delete_participant(%Participant{} = participant) do
```
oliverbarnes, line 365: better form to keep it a one-liner

lib/liquid_voting_web/schema/custom_types/uuid.ex
```diff
@@ -0,0 +1,36 @@
+defmodule LiquidVotingWeb.Schema.CustomTypes.UUID do
+  @moduledoc """
```
oliverbarnes, line 2: Interesting, this is new to me. What are the benefits of using a custom type and checking for the correct uuid format at the graphql layer (as opposed to having Ecto barf when the format is bad)?

lib/liquid_voting_web/schema/schema.ex
```diff
@@ -19,9 +20,9 @@ defmodule LiquidVotingWeb.Schema.Schema do
       resolve(&Resolvers.Voting.participants/3)
     end
 
-    @desc "Get a participant by its id"
+    @desc "Get a participant by its uuid"
     field :participant, :participant do
-      arg(:id, non_null(:id))
+      arg(:uuid, non_null(:string))
```
oliverbarnes, line 25: wouldn't it be `non_null(:uuid)` now that the UUID custom type's been imported?

oliverbarnes: for example, this custom `:datetime` type is used at the query arg level too: https://hexdocs.pm/absinthe/custom-scalars.html

```elixir
field :post, :post do
  arg :published_at, :datetime
```

lib/liquid_voting/voting.ex
```diff
@@ -237,13 +277,14 @@ defmodule LiquidVoting.Voting do
       nil
 
   """
+
```
niklaslong, line 280: Nitpick: we don't need this whitespace

lib/liquid_voting/voting.ex
```diff
@@ -254,6 +295,7 @@ defmodule LiquidVoting.Voting do
       ** (Ecto.NoResultsError)
 
   """
+
```
niklaslong, line 298: unnecessary whitespace

lib/liquid_voting_web/schema/custom_types/uuid.ex
niklaslong: Both would be fine; to me it's about better modelling the data that's passed in/out of the application.  Internally, when dealing with a `uuid`, it's of type `Ecto.UUID`. Regardless of whether we're passing in a valid `uuid` or not, it should be consistently represented and that implies casting with `Ecto.UUID.cast/1`. Thoughts?

lib/liquid_voting_web/schema/custom_types/uuid.ex
```diff
@@ -0,0 +1,36 @@
...
+    human-readable ID strings.
+    """)
+
+    serialize(&encode/1)
+    parse(&decode/1)
+  end
+
+  @spec decode(Absinthe.Blueprint.Input.String.t()) :: {:ok, term()} | :error
+  @spec decode(Absinthe.Blueprint.Input.Null.t()) :: {:ok, nil}
+  defp decode(%Absinthe.Blueprint.Input.String{value: value}) do
```
niklaslong, line 23: All three clauses of `decode` could be written as one liners

test/liquid_voting/participants_test.exs
```diff
@@ -61,8 +71,8 @@ defmodule LiquidVoting.ParticipantsTest do
       assert participant.name == @valid_attrs[:name]
     end
 
-    test "create_participant/1 with valid data creates a uuid" do
-      assert {:ok, %Participant{} = participant} = Voting.create_participant(@valid_attrs)
+    test "create_participant/1 with valid initial data creates a uuid" do
```
niklaslong, line 74: Not sure `initial` is necessary here, I think it's implied by the function's _function_ 😉 ?

jinjagit: I have created @valid_init_data, which has no uuid, for testing a uuid is created when participant created (this is different to @valid_data used in other tests (which includes a valid uuid)

lib/liquid_voting/voting.ex
jinjagit: so, we need get_participant with bang, without, or both?

lib/liquid_voting_web/schema/schema.ex
jinjagit: this looks like an oversight. I'll try it and test.

lib/liquid_voting/voting.ex
oliverbarnes: For this PR I'd just stick with what's being used currently. So if both versions are being used somewhere, keep both. If not, get rid of the zombie code

lib/liquid_voting_web/schema/custom_types/uuid.ex
oliverbarnes: Not sure. The issue of representation consistency seems a little abstract to me. Would a string representation for an integer id be cast before letting Ecto handle it?  

In practice does Ecto actually require the UUID string to be cast before finding a record with it?

Meant to confirm this myself, but am amidst `mix deps.get` conflicts here at the moment. https://stackoverflow.com/a/46799252/358518 says `Repo.get!` will try to cast the string itself.

test/liquid_voting/participants_test.exs
oliverbarnes: but I think uuid is not needed in any of the data sets, since it won't ever changed?

lib/liquid_voting_web/schema/custom_types/uuid.ex
niklaslong: In that case, we should probably decode it as a `string` and rely on `Ecto` to do the casting 🤔 
(Here are the absinthe docs on custom scalar types if anyone's curious: https://hexdocs.pm/absinthe/custom-scalars.html)

oliverbarnes: Yeah, I did check those out. I'm not against it, I like having a specific type in terms of semantics. I'd just like to understand the practical gain, if any. With the `datetime` custom scalar example it's not clear to me either.

Also good for us to understand how this will affect consumer clients. What's the behavior if the cast fails on our end? I see the `:error` being returned by `decode`, but what's the final response like? 200 OK with error fields, or a 400 Bad Request?

lib/liquid_voting/voting.ex
jinjagit: ah yes, this was a hangover from inserting debugging lines. thanks

test/liquid_voting/participants_test.exs
jinjagit: oh, that's true. I removed uuid from all @..data and it works, so I'll remove my 'special' @valid_init_data

jinjagit: This PR switches the api to using uuids, not ids, across the Participants, Votes, Delegations and Results contexts (and related db tables), as primary keys and foreign keys (where present). It covers issues #67 and #68.  

I had 2 issues related to upserting functions and the structs they return (apologies in advance for the lengthy notes):  

For reference (from before this PR's switch to uuids):  
```
def upsert_participant(attrs \\ %{}) do  
  %Participant{}  
  |> Participant.changeset(attrs)  
  |> Repo.insert(  
    on_conflict: {:replace_all_except, [:id]},  
    conflict_target: [:organization_uuid, :email]  
  )  
end  
```
part of `calculate_result!/2`:  
```
%Result{}  
|> Result.changeset(attrs)  
|> Repo.insert!(  
  on_conflict: {:replace_all_except, [:id]},  
  conflict_target: [:organization_uuid, :proposal_url]  
)  
```
(I think) the effect of an attempted `Repo.insert!` that becomes a 'replace fields' operation (an update), when using a primary key of an id type (a sequential integer), is that the db does not return an incremented id, since no new row has been created. Thus the returned struct contains the id of the row that was updated. Our uuids, however, are now autogenerated every time a new data struct is created, thus the struct returned after an upsert that just updates (rather than inserting a new row) does not contain the uuid of the updated row, rather it contains a newly generated uuid. I could be wrong about this, but it certainly seems that this is happening.  

1. Previously, when using id as primary key, `upsert_participant(%{email: <"existing_email">, organization_uuid: <valid_organization_uuid>)` would update an existing participant entry with same email, and a `%Participant{}` with the id of the existing participant would be returned. Now, a `%Participant{}` with a newly autogenerated uuid will be returned. The table row will still be updated and the uuid will not be changed in the table, but the returned uuid does not match the record in the table. Thus, I had to introduce an extra step (which sadly involves an extra db query), in `Resolvers.Voting create_vote/3`, which gets the actual uuid of a participant from the relevant db record, rather than simply using the uuid returned in the `%Participant{}` struct, after the preceeding `upsert_participant/1` call.    

2. I replaced the test in voting_results_test.exs of `calculate_result!/1` - "returns the same result struct for a given proposal_url", with; "inserts data into existing result, for a given proposal_url && organization". The old test failed using uuids. This is because it asserted the id returned by the struct was unchanged when a result for the same proposal was created (calculated) twice. This fails because of the uuid autogeneration vs sequential ids issue, described above. Thus the uuids of the structs would never match. My new test looks at what is and isn't inserted to the table instead. Please check if this actually tests what was intended to be tested. Thanks!  

I suspect there is a more elegant way to avoid this issue of the autogenerated uuids not matching after an upsert (that is an update, not an insert), or that I am misunderstanding what is happening. Ideas are welcome :)

As a side note, `upsert_participant(%{attrs})`, when `attrs` does not contain `:name "some_name`, will overwrite a previously existing name with a value of nil, which seems counter-intuitive. This seems particularly strange when considering that GraphQL(absinthe) requires a name and email for createParticipant, but only an email to upsert a participant when using createVote (which will overwrite a pre-existing name with nil, if no name provided).

Other notes:

In the GraphQL examples in the readme, I have indicated where it is necessary to insert a valid uuid from the preceding query(ies). This, unfortunately means that blindly copy/pasting all the examples will result in errors (if valid uuids are not inserted in the appropriate places).

Linting fails on GitHub with error: "The following files were not formatted: * test/liquid_voting_web/absinthe/mutations/create_participant_test.exs" I don't know why, as my last action before my last commit was `mix format`. Help would be appreciated!

lib/liquid_voting/voting.ex
jinjagit: I didn't do this... maybe move to separate issue?

oliverbarnes: yes, a separate PR at least. not worth opening an issue for it

test/liquid_voting/participants_test.exs
oliverbarnes: The detailed notes are excellent, keep them coming! Helps a lot with the review.

The un-matching uuids doesn't sound right, will have a look. And the README examples should definitely work

oliverbarnes:
Renamed the PR according to the description, for clarity

test/liquid_voting/participants_test.exs
jinjagit: Yeah, that's the only bit I don't like. My 'fix' adds a db hit too, which feels wrong.

jinjagit: Also, for clarity, the readme examples do work, IF the user pastes the correct uuid into each query/mutatioin that requires one.

README.md
```diff
@@ -131,9 +132,9 @@ query {
 }
 
 query {
-  participant(id: 1) {
+  participant(uuid: <valid_participant_uuid_string_from_previous_query_result>) {
```
oliverbarnes, line 135: ok, I see what you mean now by the query not being runnable as is. Maybe add a random UUID in these examples, though consistent if a participant is a delegate on another query etc, and add a note about needing to replace them with real ones before the examples (next to `Then run some queries:`)

test/liquid_voting/participants_test.exs
jinjagit: Maybe [Ecto.multi](https://hexdocs.pm/ecto/Ecto.Multi.html) will help with the un-matching uuids?

lib/liquid_voting/delegations/delegation.ex
```diff
@@ -4,19 +4,28 @@ defmodule LiquidVoting.Delegations.Delegation do
...
+  @primary_key {:uuid, :binary_id, autogenerate: true}
+
   schema "delegations" do
     field :proposal_url, EctoFields.URL
     field :organization_uuid, Ecto.UUID
 
-    belongs_to :delegator, Participant
-    belongs_to :delegate, Participant
+    belongs_to :delegator, Participant,
+      references: :uuid,
```
oliverbarnes, line 14: Feels like this shouldn't be needed, belongs_to ought to work out of the box with primary uuid keys. Maybe try setting `@foreign_key_type` as well as the primary key? https://hexdocs.pm/ecto/Ecto.Schema.html#module-schema-attributes

jinjagit: Good idea, I saw that in the docs, but for some reason didn't try it.

lib/liquid_voting_web/resolvers/voting.ex
```diff
@@ -40,15 +40,18 @@ defmodule LiquidVotingWeb.Resolvers.Voting do
          message: "Could not create vote with given email",
          details: ChangesetErrors.error_details(changeset)}
 
-      {:ok, participant} ->
+      {:ok, _participant} ->
+        %Voting.Participant{uuid: participant_uuid} =
```
oliverbarnes, line 44: Is there a benefit from pattern matching for `participant_uuid` here instead of using the returned `participant` in the tuple and retrieving the uuid with `participant.uuid` below?

lib/liquid_voting_web/schema/custom_types/uuid.ex
oliverbarnes: This isn't resolved :) Two questions to be answered here: what is the gain of having the custom type exactly, and how does it change the final response, if at all?

lib/liquid_voting_web/resolvers/voting.ex
jinjagit: this is the upsert issue. the uuid returned in the resulting struct does not match the uuid of the updated row in the original form (it is a newly generated uuid)

priv/repo/migrations/20200710132150_create_votes.exs
```diff
@@ -0,0 +1,27 @@
...
+      add :participant_uuid,
+          references(:participants, column: :uuid, type: :uuid, on_delete: :nothing)
+
+      add :proposal_url, :text, default: false, null: false
+      add :organization_uuid, :uuid
+
+      timestamps()
+    end
+
+    create index(:votes, [:participant_uuid])
```
oliverbarnes, line 19: this index is outdated, given all queries will include :organization_uuid now - so we might as well make it a composite index in this PR, to keep with the clean slate spirit

priv/repo/migrations/20200710132150_create_votes.exs
```diff
@@ -0,0 +1,27 @@
...
+    create index(:votes, [:proposal_url])
```
oliverbarnes, line 20: same here

priv/repo/migrations/20200710132150_create_votes.exs
jinjagit: :participant_uuid + :organization_uuid?

priv/repo/migrations/20200710132150_create_votes.exs
jinjagit: participant_uuid + :proposal_url + organization_uuid (and delete current 2 indices)?

oliverbarnes: no, because each index covers different queries - one is all votes for a participant regardless of proposal and the other all votes for a proposal regardless of participant.

so `[:participant_uuid, :organization_uuid]` for the first, and `[:proposal_ulr, :organization_uuid]` for the second one

priv/repo/migrations/20200710153620_create_delegations.exs
```diff
@@ -0,0 +1,29 @@
...
+      add :delegate_uuid,
+          references(:participants, column: :uuid, type: :uuid, on_delete: :delete_all)
+
+      add :proposal_url, :text
+      add :organization_uuid, :uuid
+
+      timestamps()
+    end
+
+    create index(:delegations, [:delegator_uuid])
```
oliverbarnes, line 20: same here and for the following two indices regarding making them composites with `:organization_uuid`

test/liquid_voting/voting_results_test.exs
```diff
@@ -26,19 +26,30 @@ defmodule LiquidVoting.VotingResultsTest do
...
-        VotingResults.calculate_result!(context[:proposal_url], context[:organization_uuid])
+      VotingResults.calculate_result!(context[:proposal_url], context[:organization_uuid])
+
+      results2 = VotingResults.list_results(context[:organization_uuid])
+
+      %Result{uuid: new_uuid, in_favor: new_in_favor} = Enum.at(results2, 0)
 
-      assert id == new_id
+      # Only one result has been added to the list of results
+      assert Enum.count(results1) == Enum.count(results2)
```
oliverbarnes, line 48: I think asserting the uuid of the returned result is still the same is enough, no need to list results.

test/liquid_voting/voting_results_test.exs
```diff
@@ -26,19 +26,30 @@ defmodule LiquidVoting.VotingResultsTest do
...
+      # The uuid of the result is unchanged
+      assert uuid == new_uuid
```
oliverbarnes, line 50: maybe `uuid_for_updated_result` instead of `new_uuid`?

test/liquid_voting_web/absinthe/mutations/create_participant_test.exs
```diff
@@ -0,0 +1,107 @@
+defmodule LiquidVotingWeb.Absinthe.Mutations.CreateParticipantTest do
```
oliverbarnes, line 1: 🙏 thanks for adding test coverage here!

README.md
jinjagit: hmm... but the consistency with the previous query will not be obvious, since we don't include the query results in the examples (though we could).

test/liquid_voting/voting_results_test.exs
jinjagit: this is the (other of the 2) upsert issues, previously described. The problem is exactly that: the uuids won't match using the test as previously was.

README.md
oliverbarnes: I see what you mean - maybe just make it a little shorter?  Like `<participant uuid fetched in previous query result>`

lib/liquid_voting_web/resolvers/voting.ex
oliverbarnes: ah, ok. I'll have to play with this to make sense of this problem. I haven't seen anything in the code that might cause something like that.

jinjagit: I'm not sure, but have a hunch the Ecto.multi operation might be the answer.

priv/repo/migrations/20200710132150_create_votes.exs
oliverbarnes: exactly 👍

README.md
jinjagit: that's longer :)

oliverbarnes:
This is looking very good, man 🦾🦾

I just have the comments above, and will play with upsert issue later today to make sense of it. I don't think Multi will necessarily fix it, Multi is great to organize complex ecto queries, but the underlying queries should still work correctly without it.

jinjagit:
Thanks. I'll get on with making the other changes suggested today, and force push them when done.

priv/repo/migrations/20200710132150_create_votes.exs
jinjagit: for clarity: all the suggested new compound indices (in the 3 migrations) are not unique indices, just indices, right?

oliverbarnes: right, they're all for finding collections of records, not a single unique one

lib/liquid_voting/delegations/delegation.ex
jinjagit: Just adding `@foreign_key_type :binary_id` and reverting to simpler:
```
belongs_to :delegator, Participant
belongs_to :delegate, Participant
```
... doesn't work. Defaults to expecting participant.id, etc.

jinjagit: works fine when I add `@foreign_key_type :binary_id` and leave my more verbose, specific, 'belongs_to' statements.

jinjagit: Didn't help to reduce the complexity of the need 'belong_to', but added '@foreign_key_type' in any case.

jinjagit:
All suggestions implemented, except:
1. Upsert uuid mismatch issue - @oliverbarnes investigating
2. Revert to simpler 'belongs_to' statements in delegation & vote schemas (adding @foriegn_key_type) did not resolve this.

lib/liquid_voting_web/resolvers/voting.ex
oliverbarnes: from our slack investigation: https://github.com/elixir-ecto/ecto/issues/1966#issuecomment-304469636

README.md
oliverbarnes: ?
```
<valid_participant_uuid_string_from_previous_query_result>
<participant uuid fetched in previous query result>
```

lib/liquid_voting/delegations/delegation.ex
oliverbarnes: try just keeping the `type: :binary_id` option but removing `references: :uuid` and `foreign_key: :delegator_uuid`? from what I've been seeing on ecto discussions, that should work

oliverbarnes:
> All suggestions implemented, except:
> 
> 1. Upsert uuid mismatch issue - @oliverbarnes investigating
> 2. Revert to simpler 'belongs_to' statements in delegation & vote schemas - adding @foriegn_key_type did not resolve this.

We found a solution for 1), and 2 is being discussed. 

There's also https://github.com/liquidvotingio/api/pull/71#discussion_r452140870 open still, too

jinjagit:
UUIDs now match following upsert. Removed unnecessary db query(ies) used as workaround.

lib/liquid_voting_web/schema/custom_types/uuid.ex
jinjagit: I don't feel qualified to comment on this.

oliverbarnes: This could be researched, but if you feel like enough effort has been put into this PR (which is understandable), one option might be to remove this change and open a separate draft PR for further discussion, to get references with time etc. No rush in adding something we don't fully understand at this point.

jinjagit: sure. I'll try removing it

jinjagit:
Custom type checking removed (I will re-include in new draft PR following merge of this PR)

lib/liquid_voting/delegations/delegation.ex
jinjagit: Tried that (now), and get: warning: invalid association `participant` in schema LiquidVoting.Voting.Vote: associated schema LiquidVoting.Voting.Participant does not have field `id`

oliverbarnes: ok, let me know if you don't have any luck with @niklaslong when you have a look together, else I'll go ahead and merge (resetting the production db just before it)

jinjagit: We just had a look together. No concerns thus far. Niklas wants to look at the PR tomorrow morning to give it a more thorough inspection.

oliverbarnes: sounds good 👍

niklaslong: We could reduce boilerplate using macros (I wouldn't bother at this point tbh as it complicates something simple) or by naming the primary keys `id` instead of `uuid` as that's what `Ecto` defaults to.

jinjagit: Yes, we could change all names that use 'uuid' (including fkeys; 'participant_uuid', for example) to 'id', and this issue would resolve. I like that we use 'uuid', however, as it highlights the use of uuids to anyone reviewing the code, db, etc. If we decide to revert to using the 'id' naming scheme, however,I can do that.

lib/liquid_voting/voting/vote.ex
```diff
@@ -2,25 +2,31 @@ defmodule LiquidVoting.Voting.Vote do
...
+  @foreign_key_type :binary_id
+
   schema "votes" do
     field :yes, :boolean, default: false
     field :weight, :integer, default: 1
     field :proposal_url, EctoFields.URL
     field :organization_uuid, Ecto.UUID
 
-    belongs_to :participant, LiquidVoting.Voting.Participant
+    belongs_to :participant, LiquidVoting.Voting.Participant,
```
niklaslong, line 14: I'd alias `LiquidVoting.Voting.Participant`

test/liquid_voting/participants_test.exs
```diff
@@ -61,7 +63,7 @@ defmodule LiquidVoting.ParticipantsTest do
       assert participant.name == @valid_attrs[:name]
     end
 
-    test "create_participant/1 with valid data creates a uuid" do
+    test "create_participant/1 with valid initial data creates a uuid" do
```
niklaslong, line 66: I'd remove the `initial` here as it's implied.

jinjagit:
p_key names changed to 'id' from 'uuid'. 
Note: @Oliver mentioned need to run db reset on production api just prior to merging this PR.

priv/repo/migrations/20200710123927_create_participants.exs
```diff
@@ -0,0 +1,20 @@
+defmodule LiquidVoting.Repo.Migrations.CreateParticipants do
+  use Ecto.Migration
+
+  def change do
+    create table(:participants, primary_key: false) do
```
niklaslong, line 5: I'm wondering if we could configure the DB to use primary keys in the `config.exs` as was shown here: https://annkissam.com/elixir/alembic/posts/2019/01/04/ecto-and-binary-ids.html?

```elixir
config :my_app, MyApp.Repo,
  migration_primary_key: [name: :id, type: :uuid]
```

jinjagit: OK, I'll try this.

PR #86, opened 2020-07-19
correct and add some @doc descriptions

jinjagit:
Just some corrections and additions to `@doc` entries in voting.ex, delegations.ex, and voting_results.ex.

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -130,7 +130,7 @@ defmodule LiquidVoting.Delegations do
   def create_delegation!(attrs \\ %{}) do
     %Delegation{}
     |> Delegation.changeset(attrs)
-    |> Repo.insert()
+    |> Repo.insert!()
   end
 
   @doc """
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -115,6 +115,18 @@ defmodule LiquidVoting.Delegations do
     |> Repo.insert()
   end
 
+  @doc """
+  Creates a delegation.
+
+  ## Examples
+
+      iex> create_delegation(%{field: value})
+      %Delegation{}
+
+      iex> create_delegation(%{field: bad_value})
+      Ecto.*Error
+
+  """
   def create_delegation!(attrs \\ %{}) do
     %Delegation{}
     |> Delegation.changeset(attrs)
--- a/lib/liquid_voting/voting_results.ex
+++ b/lib/liquid_voting/voting_results.ex
@@ -135,7 +135,7 @@ defmodule LiquidVoting.VotingResults do
   end
 
   @doc """
-  Creates a delegation.
+  Creates a result.
 
   ## Examples
 
--- a/lib/liquid_voting/voting_results.ex
+++ b/lib/liquid_voting/voting_results.ex
@@ -102,7 +102,7 @@ defmodule LiquidVoting.VotingResults do
   @doc """
   Gets a single result by its proposal url and organization_id
 
-  Raises `nil` if the Result does not exist.
+  Returns `nil` if the Result does not exist.
 
   ## Examples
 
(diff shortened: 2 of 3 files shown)


lib/liquid_voting/delegations.ex
```diff
@@ -115,6 +115,18 @@ defmodule LiquidVoting.Delegations do
...
+  @doc """
+  Creates a delegation.
+
+  ## Examples
+
+      iex> create_delegation(%{field: value})
+      %Delegation{}
+
+      iex> create_delegation(%{field: bad_value})
+      Ecto.*Error
```
niklaslong, line 127: This is great, although the function doesn't currently do what we expect as it's using `insert` without the bang (currently returns `{:error, changeset}`): 

https://github.com/liquidvotingio/api/blob/7da73d6e51bbfd7e565049efc390001edae2b07c/lib/liquid_voting/delegations.ex#L129-L135 

I've also noticed some inconstancies in the return value in our docs: sometimes it's `Ecto.*Error`, elsewhere in the codebase it's `** (Ecto.NoResultsError)`. We should be consistent and precise and use the output when running these in an `IEx` shell.

lib/liquid_voting/voting_results.ex
```diff
@@ -134,6 +134,18 @@ defmodule LiquidVoting.VotingResults do
     |> Repo.insert()
   end
 
+  @doc """
+  Creates a delegation.
```
niklaslong, line 138: This should be _result_.

lib/liquid_voting/delegations.ex
jinjagit: You mean we use different forms for the same function? Because some functions may give different error(s) to `** (Ecto.NoResultsError)`, which is why they specify `Ecto.*Error`. Or you mean sometimes the wrong error type (of the 2 mentioned) is specified for a function?

niklaslong: Ah gotcha, this looks good in that case—let's use the precise error when we know what could be returns (`get` for example) and the more general `Ecto.*error` when we can't be specific. I just wanted to make sure we were being precise with our errors.

jinjagit: regarding your first point:


api/lib/liquid_voting/delegations.ex

Lines 129 to 135 in 7da73d6
   """ 
   def create_delegation!(attrs \\ %{}) do 
     %Delegation{} 
     |> Delegation.changeset(attrs) 
     |> Repo.insert() 
   end 
  """
change the error in @do, or change to `insert!`?

niklaslong: We should have `insert!` there (there's already a clause of `create_delegation`, no bang)

niklaslong: Also you seem to be pinging quite a few people using `@`...

niklaslong: I suppose it's a marketing strategy 😂

jinjagit: I know... still working on my GH communication skills. Gimme time.

jinjagit: Made the 2 edits @niklaslong  suggested.

niklaslong:
@jinjagit could you please do a quick smoke test on the live API once the deploy job has finished running on `master`?

jinjagit:
quick smoke test = pass

lib/liquid_voting/delegations.ex
```diff
@@ -115,10 +115,22 @@ defmodule LiquidVoting.Delegations do
...
+
+  """
   def create_delegation!(attrs \\ %{}) do
     %Delegation{}
     |> Delegation.changeset(attrs)
-    |> Repo.insert()
+    |> Repo.insert!()
```
oliverbarnes, line 133: For the future - this is a needed change, but would have been better as a separate PR, because it's changing behavior and not just docs.

Issue #83, opened 2020-07-17
Add absinthe deleteParticipant action


jinjagit:
To facilitate tear-down of db data created when testing production (or other) repo. - via curl smoke tests, etc.

jinjagit:
Closed, following discussion in [#87](https://github.com/liquidvotingio/api/pull/87#issuecomment-661094441)

PR #87, opened 2020-07-19
Add delete participant absinthe action

jinjagit:
This addresses issue [#83](https://github.com/liquidvotingio/api/issues/83): "Add absinthe deleteParticipant action"

TLDR: Can now delete participants using absinthe, which facilitates tear-down of created participants when testing liquid voting api via graphQL mutations. Current solution may have unintended consequences (not necessarily bad, but maybe).

Most significant change:

In `priv/repo/migrations/20200710132150_create_votes.exs`:

From: `add :participant_id, references(:participants, on_delete: :nothing)`
To: `add :participant_id, references(:participants, on_delete: :delete_all)`

Rationale: Attempted deletion of participant with associated vote(s) raised an exception:
```
** (Ecto.ConstraintError) constraint error when attempting to delete struct:

* votes_participant_id_fkey (foreign_key_constraint)
```

We already delete delegation(s) on deletion of an associated participant, (the delegation migration specifies `on_delete: :delete_all` for both the delegator and delegate associations), thus it is logical that we also delete votes on deletion of an associated participant.

A consequence of deleting associated votes when a participant is deleted was that a subscription to a result that a deleted participant had voted on would fail to update. This is because the the associated votes are deleted at db/schema level, and hence a call to `VotingResults.publish_voting_result_change/2` was never made. I resolved this in `lib/liquid_voting_web/resolvers/voting.ex delete_participant/2` by:
1. Get participant by email
2. Create list of any votes associated with participant
3. Delete participant
4. Call `VotingResults.publish_voting_result_change/2` on all associated votes' proposal_url's
Number 2, above, also necessitated creating a new `Voting.list_votes_of_participant/2 function`, to find votes that match a participant id && organization id. I'm not sure if I should also create an index `:votes, [:participant_id, :organization_id]`, since we already have `:votes, [:organization_id, :participant_id]` (yes? because reflects order of priorities in `list_votes_of_participant/2?`).

The above seems inelegant and involves additional db hits. I suspect that there is some better way of setting up the publishing of results changes at the schema level, whenever a vote is created, updated, or deleted (including deletion due to deletion of an associated participant). It would be good to pair with someone on this.

Implications:

Although I don't see an immediate problem with automatic deletion of votes associated with a deleted participant, I can see an implicit problem with the auditing of voting results that an organization, or other party, may wish to do. For example, deleting a participant while a proposal vote is in progress is not a problem, as the result will be updated accordingly. Deleting a participant with a vote on a proposal *after* a vote is closed, however, (and after we have implemented something like 'closed' flag in voting results), will mean that a search of votes cast vs the voting result will return a discrepancy.

Indeed, this is a also the case with the deletion of votes following closing a result, in any case, regardless of whether due to the deletion of participant associated with the results, or due to direct deletion of votes (we already have a `deleteVote` action in our absinthe layer).

One solution may be to create a record of all votes deleted after a vote has closed? Or we could prevent votes from being deleted after a vote has closed (in which case I maybe should leave the votes migration association as `on_delete: :nothing`) and delete associated votes programmatically (if there is a way around the constraint error I originally encountered); = more db hits, but easier to add in deletion-prevention later (as will be decoupled from participant deletion). Thoughts?

Observations:

We have no tests of the absinthe subscription. Worth creating an issue for this?


oliverbarnes:
Great PR description, nicely contextualized and thorough on its impact👌💪

These are important points to discuss. My initial feelings is that delegations are transient and dependant on a participant existing, whereas votes should be persistent even if a participant leaves an organization. Votes should only be deleted on a deliberate un-vote action by the participant before a vote result is closed.

There will probably be exceptions to this depending on organization-specific rules (say, when a participant leaves an org mid-voting, perhaps their vote for that voting should be deleted). But I'd say the general rule is to keep the vote history intact.

At some point we might have to implement non-destructive deletion of participants, so that an associated identity is kept along with the voting history. The privacy of which will be dependant on each org's rules, and legal constraints like GDPR and right-to-be-forgotten. Something to think about in our on-going privacy debate.

For this PR I'd just remove the side-effect for deleting a participant.

oliverbarnes:
> We have no tests of the absinthe subscription. Worth creating an issue for this?

Not sure it's worth it. It involves testing things over websockets, and we might just be testing absinthe itself (which is already tested). I might have looked into it back when I first implemented it and arrived at this conclusion. But it never hurts to investigate a little, if you're curious

jinjagit:
@oliverbarnes "For this PR I'd just remove the side-effect for deleting a participant."
You mean keep votes when associated participant is deleted, right? I see the logic in that.
If yes, then any ideas on how I can resolve the original 'constraint error' I got when I naively first tried to delete the participant (before changing the association to `on_delete: :delete_all`)? Thanks.

jinjagit:
i.e. I think the child (vote) needs the reference to a participant, and will error out whenever a participant is deleted. So either the vote is associated with a participant and we cannot delete a participant without first deleting the vote (= cannot tear-down all test data created via absinthe), OR they are not associated (completely decoupled, and we then perhaps add a column in votes of 'participant_deleted?' boolean), which I guess = more db searches to find votes for a participant, etc. I could be over-thinking things here, but it feels like the association is logical, yet creates a problem with our desired persistence of votes independent of the existence of an associated participant.

oliverbarnes:
I wasn't expecting an error when `on_delete: nothing` - can you post it here, please? Vote does need a participant_id, of course, but I don't see where it would require an associated record with it. Unless `belongs_go` is actually enforcing that

jinjagit:
```
** (Ecto.ConstraintError) constraint error when attempting to delete struct:
* votes_participant_id_fkey (foreign_key_constraint)
```
as described in beginning of my 1st comment on this PR

jinjagit:
I think this also hits a deeper q. about the removal of a participant: How to audit past voting results, if there can be orphaned votes, with no associated participant? (How would one be able to check the votes were not fraudulently created in the first place?)

jinjagit:
To replicate the error: Use other repo. (where vote is still set to `on_delete: :nothing`). Create a vote. Try to delete associated participant.

oliverbarnes:
> I'm beginning to feel that participants should never be deleted, though obviously this doesn't help the (less important) tear-down issue.

Agree, I'm beginning to think the same. The history should have full integrity, unless the whole organization deletes her data. Let's close this and explore the smoke and acceptance testing teardown on another issue.

Don't feel bad about the code lost, btw - exploratory PRs are great for learning more about a domain, this has been very useful.

Issue #88, opened 2020-07-20
Teardown of data after smoke and acceptance tests


oliverbarnes:
After #87 it's clear not all data can be setup and teardown through graphql requests, so let's try another path.

One might be to follow [what we do with database migrations](https://github.com/liquidvotingio/api/blob/master/lib/liquid_voting/release.ex#L4-L12), which is to call a special release function after these external tests run.

cc @jinjagit

oliverbarnes:
These release hooks can be used with a container eval (see the prod migration runbook), and from a github action:

```
kubectl exec -ti pod/$pod \
--container api-container \
-- /opt/app/bin/liquid_voting \
eval "LiquidVoting.Release.migrate"
```

jinjagit:
So I could use something like this, with LiquidVoting.Release.teardown, for example, where teardown/0 calls my mix task?

Oliver  4 minutes ago
yes, but you wouldn't need a mix task in this case because LiquidVoting.Release already has access to the environment

jinjagit:
I was thinking it would be useful to leave as a mix task, as then it can also be used locally in dev env (although mix.ecto reset kind of achieves the same thing)

oliverbarnes:
It'd only make sense to use a mix task if you were logged-in to the server instance and mix were available to interact with the app there. In this case, a github action will be calling a remote server instance (a container in a kubernetes pod) and asking it to execute a dedicated hook that has access to the app environment. And within that hook, mix isn't available (it's already in a phoenix process), and it'd just add a lot of overhead anyway.

locally, yeah, a mix ecto.reset would suffice, since there's no problem wiping out the data. On the server of course we're only deleting data scoped to an organization id.

PR #94, opened 2020-07-24
add teardown/0 to release.ex

jinjagit:
This addresses issue [#88](https://github.com/liquidvotingio/api/issues/88)

### DIFF
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -14,7 +14,7 @@ defmodule LiquidVoting.Release do
   end
 
   @doc """
-  Deletes votes, participants and delegations for a smoke tests organization_id.
+  Deletes votes, participants and delegations for the smoke tests organization_id.
   For use before and after running smoke tests.
   """
   def teardown() do
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -2,7 +2,6 @@ defmodule LiquidVoting.Release do
   @app :liquid_voting
 
   alias LiquidVoting.{Delegations, Voting}
-  alias LiquidVoting.Voting.Participant
 
   def migrate do
     for repo <- repos() do
@@ -24,8 +23,8 @@ defmodule LiquidVoting.Release do
 
     participants = Voting.list_participants(organization_id)
 
-    Enum.each(participants, fn %Participant{} = participant ->
-      Voting.delete_participant(participant)
+    Enum.each(participants, fn participant ->
+      {:ok, participant} = Voting.delete_participant(participant)
     end)
 
     delegations = Delegations.list_delegations(organization_id)
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -14,17 +14,24 @@ defmodule LiquidVoting.Release do
   end
 
   @doc """
-  Deletes votes, participants and delegations for a specific organization_id.
+  Deletes votes, participants and delegations for a smoke tests organization_id.
   For use before and after running smoke tests.
   """
-  def teardown(organization_id) do
+  def teardown_smoke_test_data(), do: teardown("bc7eeccb-5e10-4004-8bfb-7fc68536bbd7")
+
+  @doc """
+  Deletes votes, participants and delegations for demo organization_id.
+  """
+  def teardown_demo_data(), do: teardown("62309201-d2f0-407f-875b-9f836f94f2ca")
+
+  defp teardown(organization_id) do
     votes = Voting.list_votes(organization_id)
     Enum.each(votes, fn vote -> Voting.delete_vote!(vote) end)
 
     participants = Voting.list_participants(organization_id)
 
     Enum.each(participants, fn participant ->
-      {:ok, participant} = Voting.delete_participant(participant)
+      {:ok, _participant} = Voting.delete_participant(participant)
     end)
 
     delegations = Delegations.list_delegations(organization_id)
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -1,5 +1,6 @@
 defmodule LiquidVoting.Release do
   @app :liquid_voting
+  @test_organization_id "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
 
   alias LiquidVoting.{Delegations, Voting}
 
@@ -17,19 +18,17 @@ defmodule LiquidVoting.Release do
   Deletes votes, participants and delegations for the smoke tests organization_id.
   For use before and after running smoke tests.
   """
-  def teardown() do
-    organization_id = "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
-
-    votes = Voting.list_votes(organization_id)
+  def smoke_test_teardown() do
+    votes = Voting.list_votes(@test_organization_id)
     Enum.each(votes, fn vote -> Voting.delete_vote!(vote) end)
 
-    participants = Voting.list_participants(organization_id)
+    participants = Voting.list_participants(@test_organization_id)
 
     Enum.each(participants, fn participant ->
       {:ok, _participant} = Voting.delete_participant(participant)
     end)
 
-    delegations = Delegations.list_delegations(organization_id)
+    delegations = Delegations.list_delegations(@test_organization_id)
     Enum.each(delegations, fn delegation -> Delegations.delete_delegation!(delegation) end)
   end
 
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -17,7 +17,6 @@ defmodule LiquidVoting.Release do
   Deletes votes, participants and delegations for a smoke tests organization_id.
   For use before and after running smoke tests.
   """
-
   def teardown() do
     organization_id = "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
 
--- a/lib/liquid_voting/release.ex
+++ b/lib/liquid_voting/release.ex
@@ -19,6 +19,16 @@ defmodule LiquidVoting.Release do
   For use before and after running smoke tests.
   """
   def smoke_test_teardown() do
+    for repo <- repos() do
+      {:ok, _, _} =
+        Ecto.Migrator.with_repo(repo, fn repo ->
+          Ecto.Migrator.run(repo, :up, all: true)
+          run_teardown()
+        end)
+    end
+  end
+
+  defp run_teardown() do
     @test_organization_id
     |> Voting.list_votes()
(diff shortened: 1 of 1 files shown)


jinjagit:
Would we prefer this def to have organization_id accepted as arg? e.g. call teardown("bc7eeccb-5e10-4004-8bfb-7fc68536bbd7")

lib/liquid_voting/release.ex
```diff
@@ -11,6 +13,23 @@ defmodule LiquidVoting.Release do
...
+  For use before and after running smoke tests.
+  """
+  def teardown() do
+    organization_id = "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
+
+    votes = Voting.list_votes(organization_id)
+    Enum.each(votes, fn vote -> Voting.delete_vote!(vote) end)
+
+    participants = Voting.list_participants(organization_id)
+    Enum.each(participants, fn participant -> Voting.delete_participant(participant) end)
```
niklaslong, line 27: We should probably use the `!` version of this (or pattern match on the response at the very least)?

niklaslong, line 27: We should probably use the `!` version of this (or pattern match on the response at the very least)?

jinjagit: I guess pattern match then, since there is no ! version

niklaslong:
I'm in favour of passing in the `organisation_id`; that way we're not dependant on a hardcoded value.

lib/liquid_voting/release.ex
jinjagit: Not sure I unsderstand the use of pattern matching the response, since it will be a list (empty, or with contents).

jinjagit: Added pattern matching for return value of calls to delete_participant/1, and organization_id is now passed in to the def.

oliverbarnes:
The benefit of sticking to a hardcoded value would be that teardown can only be called on that given org, even if a malicious user somehow gains access to the release app (or to the smoke tests themselves). 

Then again, if someone gains access to the release they're also able to run `rollback` 😅, so that scenario may be moot. 

Mistakes on the smoke test scripts may still be an issue? Like, pasting the wrong org id by mistake and deleting a whole real organization's data in the process.... For that, the developer would need to have an organization id in their clipboard. From, say, browsing the production auth database.

Perhaps both the demo org id and test org ids should be permanent and hard coded, is what I'm saying. And better not be flexible with them. What do you think?

jinjagit:
I agree. Hardcoding is best , to prevent accidental deletion (I went on the same mental journey as you, Oliver, re. malicious use ;-) )

lib/liquid_voting/release.ex
niklaslong: It's to throw an error if something goes wrong, otherwise we would have no way of knowing.

jinjagit: Yeah, eventually I worked it out, I think.

niklaslong:
True, those reasons definitely outweigh the benefits of not having it hardcoded. 👍

oliverbarnes:
Cool 👍 

> I went on the same mental detour as you, Oliver, re. malicious use ;-)

malicious use speculation always fun (and labyrinthic) 

> I'll hardcode the test and demo ids, then we can pass a value in to switch to either - or some similar idea.

Actually, I don't mean hardcoding both demo and test orgs id here - we just need the test org id in the teardown function. I mean the demo org id should likewise be permanent and hardcoded where relevant (docs).

jinjagit:
organization_id value now hard-coded into teardown/0

lib/liquid_voting/release.ex
```diff
@@ -11,6 +13,26 @@ defmodule LiquidVoting.Release do
     {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :down, to: version))
   end
 
+  @doc """
+  Deletes votes, participants and delegations for the smoke tests organization_id.
+  For use before and after running smoke tests.
+  """
+  def teardown() do
+    organization_id = "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
```
oliverbarnes, line 21: it's good practice to define constants outside of functions, to keep them nice and visible when browsing the source. Here you could extract this and define it as a module attribute like this:

```
defmodule LiquidVoting.Release do
  @app :liquid_voting
  @test_organization_id "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
```

(prepending _test__ makes it more explicit too)

lib/liquid_voting/release.ex
```diff
@@ -11,6 +13,26 @@ defmodule LiquidVoting.Release do
     {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :down, to: version))
   end
 
+  @doc """
+  Deletes votes, participants and delegations for the smoke tests organization_id.
+  For use before and after running smoke tests.
+  """
+  def teardown() do
```
oliverbarnes, line 20: wondering if `smoke_test_teardown()` would be clearer

jinjagit:
I think this ready to merge now.

lib/liquid_voting/release.ex
```diff
@@ -11,6 +13,26 @@ defmodule LiquidVoting.Release do
...
+    organization_id = "bc7eeccb-5e10-4004-8bfb-7fc68536bbd7"
+
+    votes = Voting.list_votes(organization_id)
+    Enum.each(votes, fn vote -> Voting.delete_vote!(vote) end)
```
oliverbarnes, line 24: Maybe 

```
organization_id
|> Voting.list_votes
|> Enum.each(fn vote -> Voting.delete_vote!(vote) end)
```

@niklaslong what do you think?

oliverbarnes: and if so, same for participants and delegations

jinjagit: function renamed, and organization_id set as module attribute

jinjagit: refactored to use pipe construction

jinjagit:
FYI... the mix task equivalent included `import Ecto.Query, warn: false`, and also ran the app with: `Mix.Task.run("app.start", [])`

oliverbarnes:
> FYI... the mix task equivalent includes `import Ecto.Query, warn: false`, and also runs the app with: `Mix.Task.run("app.start", [])`

mix & releases are different beasts. it's a good point to get familiar with them:

https://hexdocs.pm/phoenix/releases.html
https://hexdocs.pm/mix/Mix.Tasks.Release.html

jinjagit:
Yes, i mean maybe we need to start the app, etc. in the release code. E.g.
```
# To start your system
    _build/prod/rel/my_app/bin/my_app start
```

niklaslong:
I think we'll need something similar to what is implemented for `Ecto.Migrator.with_repo/3`, i.e. ensuring repo is started and if not start manually. 

https://github.com/elixir-ecto/ecto_sql/blob/2e1378eb5d65c2f34d49b595891e7ab482b44507/lib/ecto/migrator.ex#L124

jinjagit:
This appears to be doing something now.
`eval "LiquidVoting.Release.smoke_test_teardown"` now returns `12:38:29.673 [info] Already up`

Since I have only a vague idea of what I am really doing, here is what I did / how I tested it:

1. edited release.ex using Niklas's observation, in previous comment.
2. started production locally using Oliver's post with relevant commands
3. opened liquid_voting_dev on psql and checked participants table is empty
4. started api server locally
5. ran k6 delegations_test
6. checked participants table now included new participants (true)
7. ran the eval command (and preceding code block), mentioned at top of this comment.
8. checked participants table now empty (true)

niklaslong:
It does seem like progress! 💪 
I don't think we actually need to run a migration though. How about:
```elixir
def smoke_test_teardown() do
  for repo <- repos() do
    {:ok, _, _} = Ecto.Migrator.with_repo(repo, fn _repo -> run_teardown() end)
  end
end
```

jinjagit:
> It does seem like progress! muscle
> I don't think we actually need to run a migration though. How about:
> 
> ```elixir
> def smoke_test_teardown() do
>   for repo <- repos() do
>     {:ok, _, _} = Ecto.Migrator.with_repo(repo, fn _repo -> run_teardown() end)
>   end
> end
> ```

Nice. I was wondering about that. Seems to work fine without migration.

lib/liquid_voting/release.ex
```diff
@@ -11,6 +14,32 @@ defmodule LiquidVoting.Release do
     {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :down, to: version))
   end
 
+  @doc """
+  Deletes votes, participants and delegations for the smoke tests organization_id.
+  For use before and after running smoke tests.
+  """
+  def smoke_test_teardown() do
+    for repo <- repos() do
+      {:ok, _, _} = Ecto.Migrator.with_repo(repo, fn _repo -> run_teardown() end)
```
oliverbarnes, line 23: 👏

lib/liquid_voting/release.ex
```diff
@@ -11,6 +14,32 @@ defmodule LiquidVoting.Release do
...
+    end
+  end
+
+  defp run_teardown() do
+    @test_organization_id
```
oliverbarnes, line 28: It'd be good to have some output to sure we know this ran sucessfully. Maybe a basic count of resources before the teardown runs, and another after? Output could be:

```
## About to run smoke test teardown 
Current resource counts: 
10 Participants, 5 Votes, 3 Delegations
## Running smoke test teardown on test organization data
## Teardown ran successfully.
End resource counts: 
0 Participants, 0 Votes, 0 Delegations
## Have a nice day
```

jinjagit: sure.

jinjagit: After running all 4 smoke tests, `eval "LiquidVoting.Release.smoke_test_teardown"` now outputs: 
teardown-message

jinjagit:
Is the organization of code / defs OK?

oliverbarnes:
> Is the organization of code / defs OK?

Yeah the printing of resources being a separate function separates concerns just enough, keeping things DRY. It does run the same queries for listing resources 3 times instead of twice (the initial listing could be re-used for the deletion after the count printout), but I wasn't too concerned about that - these will be very quick queries on a tiny number of rows each time.

jinjagit:
Cool. Thanks for checking.

lib/liquid_voting/release.ex
oliverbarnes: 🍭 sweet

niklaslong: Love the polite code ❤️

Issue #78, opened 2020-07-16
Change use of org-uuid to org-id in headers


jinjagit:
When [#28](https://github.com/liquidvotingio/auth/issues/28) (auth: "Change p_key name to 'id' from 'uuid') is done, it makes sense to use 'id' in headers, rather than uuid.

Edit of @oliverbarnes comment re. this:

These are dependent on the mapping done on https://github.com/liquidvotingio/api/blob/master/lib/liquid_voting_web/plugs/context.ex#L17-L19

  def build_context(conn) do
    with [organization_uuid] <- get_req_header(conn, "org-uuid") do
      %{organization_uuid: organization_uuid}

Updating the headers will have to be coordinated across different repos though, not only api and auth, but ruby-client and deployment as well - the kubernetes ingress configures the header the auth forwards to api: https://github.com/liquidvotingio/deployment/blob/master/k8s/api-ingress.yaml#L8

nginx.ingress.kubernetes.io/auth-response-headers: Org-UUID, X-Request-ID

niklaslong:
Current master has: https://github.com/liquidvotingio/api/blob/master/lib/liquid_voting_web/plugs/context.ex#L17-23

jinjagit:
Also:
```
auth/lib/liquid_voting_auth_web/controllers/auth_controller.ex

// lines 11-15
if org = registered?(auth_key) do
        conn
        |> put_resp_header("org-uuid", org.id)
        |> send_resp(200, "")
      else
```
```
auth/test/liquid_voting_auth_web/controllers/auth_controller_test.exs

// line 28
assert Plug.Conn.get_resp_header(conn, "org-uuid") == [organization.id]
```
No results (relevant to this issue) for 'uuid' search: 
- Ruby client
- liquid-voting-website
- admin
- LiquidVoting Browser Extension

Not searched:
- decidim
- decidim barcelona
- paticipa2

jinjagit:
See PRs: [#101](https://github.com/liquidvotingio/api/pull/101), ...

PR #101, opened 2020-07-31
change Org-UUID to Org-ID in headers

jinjagit:
This is part of [#78](https://github.com/liquidvotingio/api/issues/78)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,10 +65,10 @@ docker run -it --rm \
 
 ### Once you're up and running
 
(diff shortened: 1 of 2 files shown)


jinjagit:
> README also has a reference to Org-UUID:
> 
> ```
> Once you're up and running
> Open a GraphiQL window in your browser at http://localhost:4000/graphiql, then configure a Org-UUID header:
> 
> { "Org-UUID": "b7a9cae5-6e3a-48b1-8730-8b5c8d6c9b5a"}
> ```

good catch... and done

PR #103, opened 2020-08-04
WIP - add-sobelow-security-scan

jinjagit:


jinjagit:
WIP - Current status:  

CI fails if sobelow scan returns any issues, and passes if returns zero issues.  

Method: `grep` for "Confidence" in sobelow output, since this word should exist in any reported issue, and assign number of incidences found to env var. Fail CI using `exit 1` if env var value != 0.

Issue #92, opened 2020-07-23
Upsert conflicting/duplicate delegations by same delegator


jinjagit:
At present, it is possible to create 2 or more delegations for the same delegator (global and/or proposal specific). This will add an extra vote per delegation (not desirable beyond the first delegation).  

We need therefore, to validate "delegator is not already the delegator for a similar delegation (global or proposal-specific)' before creating a delegation. This validation should also prevent creation of proposal specific delegations for a delegator who is already a global delegation delegator.

oliverbarnes:
Edited title to be more to the point. Validations will be the implementation detail

oliverbarnes:
So, being specific we want to prevent:

- a proposal-specific delegation from being created if there's already a global one with the same delegator and delegate
- more than one global delegation per delegator
- more than one proposal-specific delegation per delegator and proposal

Is this right? If so, we probably actually need separate issues for each too =D

oliverbarnes:
Related: https://github.com/liquidvotingio/api/issues/24

jinjagit:
Oh. I just realized we also cannot create 2 delegations for different proposals, with same delegator / delegate pairing.

oliverbarnes:
Ok, so 

- ~~proposal-specific delegation from being created if there's already a global one with the same delegator and delegate~~

Regarding 👇 

> that a delegator can make a global and specific delegations to different delegates (and this will add an extra vote to totals, if both delegates votes are cast)...

As per https://github.com/liquidvotingio/api/issues/24 (to be implemented still), proposal-specific delegations to different delegates should override global delegations. So when both delegates vote, only the vote from the proposal-specific delegate counts.

jinjagit:
Agree... I think! When [#24](https://github.com/liquidvotingio/api/issues/24) is implemented, then 2nd and 3rd bullet points close the holes I was concerned with here.

jinjagit:
Status: We are currently reconsidering this approach, and wondering if we should be updating rather than preventing the types of delegations mentioned in the related issue. We are also considering whether a global delegation boolean flag would be helpful. (copied from PR [#107](https://github.com/liquidvotingio/api/pull/107))

PR #107, opened 2020-08-11
WIP - Prevent conflicting/duplicate delegations by same delegator

jinjagit:
Closes [#92](https://github.com/liquidvotingio/api/issues/92)


jinjagit:
Status: We are currently reconsidering this approach, and wondering if we should be updating rather than preventing the types of delegations mentioned in the related issue. We are also considering whether a global delegation boolean flag would be helpful.

jinjagit:
Closing this, as I am now working on a new branch for this issue.

PR #111, opened 2020-08-17
Add and set global boolean for delegations

jinjagit:
Closes [#110](https://github.com/liquidvotingio/api/issues/110)  

I have not exposed the `global` boolean to the absinthe layer as I was not sure if we want to do this, and it also seems beyond the scope of this issue. If we do expose this field to the absinthe layer, then we could list delegations by `global == true` or `global == false` via GraphQL queries, for example. Thoughts?

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -116,8 +116,6 @@ defmodule LiquidVoting.Delegations do
   def create_delegation(attrs \\ %{})
 
   def create_delegation(%{delegator_email: _, delegate_email: _} = args) do
-    require IEx
-    IEx.pry()
     delegator_attrs = %{email: args.delegator_email, organization_id: args.organization_id}
     delegate_attrs = %{email: args.delegate_email, organization_id: args.organization_id}
     attrs = Map.take(args, [:organization_id, :proposal_url])
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -101,8 +101,8 @@ defmodule LiquidVoting.Delegations do
   @doc """
   Creates a delegation.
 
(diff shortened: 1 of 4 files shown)


lib/liquid_voting/delegations.ex
```diff
@@ -101,8 +101,8 @@ defmodule LiquidVoting.Delegations do
   @doc """
   Creates a delegation.
 
-  The delegation will be global if no `proposal_url` is passed in. 
-  The delegation can be created by ID or by email. 
+  The delegation will be global if no `proposal_url` is passed in.
```
jinjagit, line 104: I think the changes to this file are just white-space deletion when running `mix.format`.

test/liquid_voting/delegations_test.exs
```diff
@@ -51,7 +51,7 @@ defmodule LiquidVoting.DelegationsTest do
       assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(context[:invalid_attrs])
     end
 
-    test "create_delegation/1 with proposal urls creates a delegation", context do
+    test "create_delegation/1 with proposal url creates a delegation", context do
```
niklaslong, line 54: 👍

jinjagit:
> Have you done a manual check to make sure behaviour remains unchanged (I'd be surprised but good to check)?

Yes image

oliverbarnes:
👌 great work, dude!

> I have not exposed the global boolean to the absinthe layer as I was not sure if we want to do this, and it also seems beyond the scope of this issue. If we do expose this field to the absinthe layer, then we could list delegations by global == true or global == false via GraphQL queries, for example. Thoughts?

Yeah, it's definitely beyond the scope here. And it might be worth opening an issue for it to discuss

PR #112, opened 2020-08-17
Upsert conflicting duplicate delegations by same delegator

jinjagit:
Closes [#92](https://github.com/liquidvotingio/api/issues/92)


jinjagit:
Current status:

Still attempting to get a basic `upsert_delegation/1` function to work.

Delegations with same delegator and same proposal url WILL update via `upsert_delegation/1`, but global delegations for the same delegator will NOT. The former was simply achieved using `:uniq_index_org_delegator_proposal` and a related `on_conflict` statement. Unfortunately, although I can create a partial index `:uniq_index_org_delegator_global_where_global_true` which _will_ prevent further global delegation creation with the same delegator, I cannot leverage it to use an `on_conflict` statement for upserting.

This limitation has been discussed in this [github thread](https://github.com/elixir-ecto/ecto/issues/2081) and this [elixir forum thread](https://elixirforum.com/t/how-to-upsert-on-unique-partial-index-postgres/28977/1). I cannot find any solution(s).

I note that although the examples in the thread use `conflict_target: {:constraint, :uidx_metrics_user_date_type}`, and my code currently uses `conflict_target: [:organization_id, :delegator_id, :global]`, I have tried the syntax used in the examples and got the same error as is reported by those examples, namely: 
```
** (Postgrex.Error) ERROR 42704 (undefined_object) constraint "uniq_index_org_delegator_global_where_global_true" for table "delegations" does not exist
```
In its current form (mirroring the `on_conflict` syntax that works fine for the proposal-specific unique index), the error is:
```
** (Postgrex.Error) ERROR 42P10 (invalid_column_reference) there is no unique or exclusion constraint matching the ON CONFLICT specification
```   
So, it appears that we aren't going to be using a partial index (one that uses `where global == true`, for example) to solve our upsert issues. This may well also mean that the global boolean flag is not needed.  
I suspect that we may need to use a different method for global delegations for the same delegator. Thoughts?

jinjagit:
Status update:  
Basic `Delegations.upsert_delegation/1` now working. Using 2 different `on_conflict `actions within `Repo.insert`, chosen conditionally on value of `changeset.global == "is_global" || ==nil` (pseudocode), with the former condition using `:uniq_index_org_delegator_global`, and the latter using `:uniq_index_org_delegator_proposal` as the conflict target.  
Next step: Decide where/when to make use of this upsert function.

oliverbarnes:
Good progress, man 👌 thanks for the detailed and clear update.

jinjagit:
###Notes:

1. We need to use a `:global = nil` value for not-global, to create the unique index `:uniq_index_org_delegator_global` that only includes global delegations. We cannot use a partial index using `where`, (for example; `:organization_id, :delegator, :global, where 'global == true'`), since partial indices do not work with the `on_conflict` action we wish to include in `Repo.insert`.  
2. I changed the `:global` field to a string, which is either `nil` or `'is_global'`, as this is less confusing than using a boolean but only ever setting `true` (vs. `nil`).  
3. The upsert `on_conflict` statements use `replace_all` without `returning: true` which means a delegation that is updated via upsert will have a new `id` inserted, which seems logical and also will enable easier recording of 'deleted' delegations should we wish to track such changes in future. (This probably necessitates a front-end update for a page showing the upserted delegation, but this would probably be the case whether the id changed or not, since something other than the id has also changed - e.g. the delegate.)  
4. The pre-existing `:uniq_index_org_delegator_delegate` prevents testing the new upsert function properly creates a new proposal-specific delegation for the same delegator & delegate where a proposal-specific delegation already exists, but with a different `proposal_url`. This is covered by another issue, and therefore out of this PR's scope, but I am confident it will work (probably as simple as just removing the index).  
4. The `LV.Delegations.update_delegation/2` will update a global -> proposal-specific delegation with no problems, but to update a proposal-specific -> global delegation it is necessary to provide `proposal_url: nil` in the args (as otherwise the original proposal_url will persist). I added a test, with an explanatory comment, to help clarify / remind of this point. This behavior seems logical, and I note we have no updateDelegation action in the absinthe layer, in any case.  
5. The `:global` field has not been exposed to the absinthe layer (not in absinthe schema, etc.). This could be covered in another issue if desired?  
6. The 2nd create_delegation/1 function in LV.Delegations:  
```
def create_delegation(attrs) do
    %Delegation{}
    |> Delegation.changeset(attrs)
    |> Repo.insert()
  end
```
is no longer used by the absinthe layer. Zombie function, or still useful?

jinjagit:
#Explaining the global field and related index:

1. We have a `uniq_index_org_delegator_proposal` which indexes delegations for a specific `proposal_url` / delegator /org  
2. this index will not include records where `proposal_url = nil`  
3. We can use this index in an `on_conflict` to upsert (update) if a new proposal-specific delegation is created for a delegator who already has an existing delegation for the same proposal == what we want to happen.  
4. We cannot use this index to similarly upsert a global delegation (to update an existing delegator global delegation to different delegate), since the index does not include the original global delegation, as this is the `proposal_url = nil` case, mentioned in 2. above.  
5. Thus, we create a `uniq_index_org_delegator_global` and set :global = 'is_global' in changeset if `proposal_url = nil` in changeset.  
6. Thus we now can use 'on_conflict' against this index to upsert (update) when we create a new global delegation for a delegator who has an existing global delegation == what we want.  
7. We then choose which on_conflict to use depending on the value of :global passed to Repo.insert.  
8. The issue of needing to use `put_change(changeset, :global, nil)` in the changeset for a non-global case, (rather than 'trust' it will never need setting as such) is somewhat of a side-issue related to the update_delegation/2 function potentially producing a record with a proposal_url = "some_url" AND a global = 'is_global', which would be wrong. I note that the absinthe layer does not make use of update_delegation/2 at them moment, however.

jinjagit:
Change of approach:

Following a dev meeting today, we decided to address the issue of duplicate delegations (for same delegator) via explicit business logic within relevant CRUD functions (for example, `create_delegation`, `upsert_delegation`), rather than via the `on_conflict` based upserts (which leverage 2 different unique indices applied as constraints in the changeset).  
This may be facilitated via the use of simpler `org_delegator` index, for example. This may also permit removal of the recently created `:global` field,

lib/liquid_voting/delegations.ex
```diff
@@ -142,6 +142,48 @@ defmodule LiquidVoting.Delegations do
...
+      |> where_proposal(proposal_url)
+      |> Repo.one()
+
+    upsert(conflicting_delegation, attrs)
+  end
+
+  defp get_or_set_proposal(%{proposal_url: proposal_url}), do: proposal_url
+  defp get_or_set_proposal(_attrs), do: nil
+
+  defp where_proposal(query, _proposal_url = nil),
```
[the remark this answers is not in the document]
niklaslong: Here's an example of what I did last time I had to build something similar: 
https://github.com/simplabs/breethe-server/blob/master/apps/breethe/lib/breethe/data/measurement.ex

This of course is a matter of personal preference, I just wanted to suggest it.

lib/liquid_voting/delegations.ex
```diff
@@ -142,6 +142,48 @@ defmodule LiquidVoting.Delegations do
...
+    do: query |> where([d], is_nil(d.proposal_url))
+
+  defp where_proposal(query, proposal_url),
+    do: query |> where([d], d.proposal_url == ^proposal_url)
+
+  defp upsert(_delegation = nil, attrs), do: create_delegation(attrs)
+  defp upsert(delegation, attrs), do: update_delegation(delegation, attrs)
```
niklaslong, line 185: I wonder if we could replace these with `Repo.insert_or_update/2`?
https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_or_update/2?

test/liquid_voting/delegations_test.exs
```diff
@@ -61,24 +64,36 @@ defmodule LiquidVoting.DelegationsTest do
...
+    test "upsert_delegation/1 with duplicate delegator and proposal_url updates the respective delegation",
+         context do
+      args = Map.merge(context[:valid_attrs], %{proposal_url: @proposal_url})
+      {:ok, %Delegation{} = delegation1} = Delegations.create_delegation(args)
 
-      assert delegation.global == false
+      args = Map.merge(context[:update_attrs], %{proposal_url: @proposal_url})
+
+      assert {:ok, %Delegation{} = delegation2} = Delegations.upsert_delegation(args)
+      assert delegation1.organization_id == delegation2.organization_id
```
niklaslong, line 80:
```suggestion
      assert delegation_1.organization_id == delegation_2.organization_id
```
naming_suggestion: `delegation` and `modified_delegation` or something a little more descriptive than `1` and `2`?

test/liquid_voting/delegations_test.exs
```diff
@@ -61,24 +64,36 @@ defmodule LiquidVoting.DelegationsTest do
...
-      assert delegation.global == true
-    end
+      assert {:ok, %Delegation{} = delegation2} =
+               Delegations.upsert_delegation(context[:update_attrs])
 
-    test "create_delegation/1 with duplicate data returns error changeset", context do
-      Delegations.create_delegation(context[:valid_attrs])
-      assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(context[:valid_attrs])
+      assert delegation1.organization_id == delegation2.organization_id
+      assert Enum.count(Delegations.list_delegations(delegation1.organization_id)) == 1
```
niklaslong, line 94: Perhaps a pipe here would be preferable?

lib/liquid_voting/delegations.ex
```diff
@@ -142,6 +142,48 @@ defmodule LiquidVoting.Delegations do
...
+    %{delegator_id: delegator_id} = attrs
+    proposal_url = get_or_set_proposal(attrs)
+
+    conflicting_delegation =
+      Delegation
+      |> where([d], d.delegator_id == ^delegator_id)
+      |> where_proposal(proposal_url)
+      |> Repo.one()
+
+    upsert(conflicting_delegation, attrs)
```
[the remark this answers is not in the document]
jinjagit: What kind of comment?

oliverbarnes:
Looking great 💪 I gave it one pass and will wait when Niklas is done with his review and changes are applied.

(btw, PR's still set to draft and title says WIP)

lib/liquid_voting/delegations.ex
```diff
@@ -142,6 +142,48 @@ defmodule LiquidVoting.Delegations do
...
+
+      iex> upsert_delegation(%{field: value})
+      {:ok, %Delegation{}}
+
+      iex> upsert_delegation(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+  """
+  def upsert_delegation(attrs \\ %{}) do
+    %{delegator_id: delegator_id} = attrs
+    proposal_url = get_or_set_proposal(attrs)
```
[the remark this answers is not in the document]
jinjagit: I get that this syntax does the same thing, but why is it preferable?

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
jinjagit: I guess my thinking is that they are only used locally to where they are. If we called them from elsewhere as well, then moving them to Delegation makes more sense to me.

lib/liquid_voting/delegations.ex
jinjagit: Interesting. Didn't know that is a thing. Will try.

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
niklaslong: It's shorter (and from the std lib) and the logic is right there rather than in functions lower down. If you think this code is too opaque, you could make the default explicit as being `nil` in the third argument, though it is the default for the function and so probably unnecessary.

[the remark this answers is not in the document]
niklaslong: Also, I didn't understand `get_or_set_proposal` before reading the code, I don't think a named function adds clarity.

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
niklaslong: Something like `# Check for conflicting delegation` 🤷 
Basically moving the information given in the variable name to a comment.

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
jinjagit: But my approach adds `proposal_url` key with value `nil` too `attrs` if not present in `attrs`. Would the approach you suggest do that too?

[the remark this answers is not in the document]
jinjagit: actually, your approach works fine, and I can't remember why I thought I need the more complex approach.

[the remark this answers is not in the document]
niklaslong: It's doing exactly the same thing 😉

lib/liquid_voting/delegations.ex
jinjagit: This will take me a while, as the example uses:   
```
result =
  case MyRepo.get(Post, id) do
    nil  -> %Post{id: id} # Post not found, we build one
    post -> post          # Post exists, let's use it
  end
  |> Post.changeset(changes)
  |> MyRepo.insert_or_update
```
So, I will need to find a way to wrap my query in a case statement, etc.

niklaslong: I think you can pipe into it:

```elixir
Delegation
      |> where([d], d.delegator_id == ^delegator_id)
      |> where_proposal(proposal_url)
      |> Repo.one()
      |> case do
          nil -> ...
          delegation -> ...
      end
      |> ....

Though in this case, it might be preferable to keep the case logic in the name functions (again personal preference based on how easy it is to understand the code).
```

jinjagit: OK, but where do I get the `changes` var from in the example?

jinjagit: Also, I have different queries, based on whether proposal_url is nil.

jinjagit: Ah, wait, that would be `attrs` right?

jinjagit: Seems to work!! :)

Should I add the...
```
case result do
  {:ok, struct}       -> # Inserted or updated with success
  {:error, changeset} -> # Something went wrong
```
?

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
jinjagit: irrelevant now, as your later `proposal_url = Map.get(attrs, :proposal_url)` removed these :)

lib/liquid_voting/delegations.ex
niklaslong: I don't think there's a need to match on the result as we don't have any custom error handling for now (other than at the GraphQL layer)

jinjagit:
All suggested changes addressed

oliverbarnes:
Cool, 👀 ing

oliverbarnes:
I just saw another commit get pushed - is this ready for another review?

jinjagit:
> I just saw another commit get pushed - is this ready for another review?

yup... just tweaking the @doc... nothing mission critical.

oliverbarnes:
Please always do a total freeze, even with docs, before asking for a review. Else we run into review comments on stale stuff, confusion, etc.

oliverbarnes:
> Please always do a total freeze, even with docs, before asking for a review. Else we run into review comments on stale stuff, confusion ensues, etc.

jinjagit:
OK, sorry.

test/liquid_voting/delegations_test.exs
```diff
@@ -61,24 +64,47 @@ defmodule LiquidVoting.DelegationsTest do
...
+      Delegations.create_delegation(context[:valid_attrs])
+      assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(context[:valid_attrs])
+    end
 
-      args = Map.merge(context[:valid_attrs], %{proposal_url: proposal_url})
-      {:ok, %Delegation{} = delegation} = Delegations.create_delegation(args)
+    test "upsert_delegation/1 with duplicate delegator and proposal_url updates the respective delegation",
+         context do
+      args = Map.merge(context[:valid_attrs], %{proposal_url: @proposal_url})
+      {:ok, %Delegation{} = original_delegation} = Delegations.create_delegation(args)
```
oliverbarnes, line 75: Better to use a [factory](https://github.com/thoughtbot/ex_machina) here. This separates setup better from code under test, and code will be more concise. Also here specifically we avoid another write to the db - `context[:valid_attrs]` has values from another factory inserted on `setup()` above.

Factories can have custom fields too, as in `insert(:delegation, proposal_url: '...')`

test/liquid_voting/delegations_test.exs
```diff
@@ -61,24 +64,47 @@ defmodule LiquidVoting.DelegationsTest do
...
+      assert original_delegation.organization_id == modified_delegation.organization_id
+      assert original_delegation.delegate_id != modified_delegation.delegate_id
+      assert original_delegation.delegator_id == modified_delegation.delegator_id
+
+      assert(
+        Delegations.list_delegations(original_delegation.organization_id)
+        |> Enum.count() ==
+          1
+      )
     end
```
oliverbarnes, line 109: Would be great to have scenarios for delegation creation with upserts, as well, to complement the update scenarios above.

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
```diff
@@ -6,7 +6,9 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegationTest do
 
   @new_delegator_email "new-delegator@email.com"
   @new_delegate_email "new-delegate@email.com"
+  @new_other_delegate_email "new-other-delegate@email.com"
```
oliverbarnes, line 9: Might be worth renaming these variables now to improve legibility - I'd remove the `@new_` prepending all of them, so we're left with `@delegator_email`, `@delegate_email`, and `@another_delegate_email`

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
```diff
@@ -191,4 +193,129 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegationTest do
...
+        delegations {
+          id 
+        }
+      }
+      """
+
+      {:ok, %{data: %{"delegations" => delegations}}} =
+        Absinthe.run(query, Schema, context: %{organization_id: @organization_id})
+
+      assert Enum.count(delegations) == 1
```
oliverbarnes, line 256: I don't think we need to test the delegation count here, since we're already testing it on the LiquidVoting.Delegations layer.

Graphql tests (and most API layer tests) are really just to meant to assert what a client consuming the API should expect. Database counts aren't their concern, unless they're part of the payload being sent.

oliverbarnes: One thing that could be tested as an API, but I'm not too sure at this point, is sending out a query for the previous delegation, to assert it doesn't exist anymore and the client doesn't get anything back. What do you think?

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
```diff
@@ -191,4 +193,129 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegationTest do
...
+        delegations {
+          id 
+        }
+      }
+      """
+
+      {:ok, %{data: %{"delegations" => [delegations]}}} =
+        Absinthe.run(query, Schema, context: %{organization_id: @organization_id})
+
+      assert Enum.count(delegations) == 1
```
oliverbarnes, line 318: Same comment and question as for the previous scenario.

oliverbarnes (changes requested):
This is excellent, @jinjagit 👌 I requested a few changes, let me know if anything doesn't make sense

oliverbarnes:
Added the PR to the Kanban board.

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
jinjagit: The previous delegation exists in modified form. Thus, the client would get the modified delegation back (with a query using `delegationId`).

lib/liquid_voting/delegations.ex
```diff
@@ -142,6 +142,47 @@ defmodule LiquidVoting.Delegations do
...
+
+  ## Examples
+
+      iex> upsert_delegation(%{field: value})
+      {:ok, %Delegation{}}
+
+      iex> upsert_delegation(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+  """
+  def upsert_delegation(attrs \\ %{}) do
```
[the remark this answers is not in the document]
jinjagit: `Delegations.create_delegation(%{})` returns an invalid changeset, with appropriate errors:
```
{:error,
 #Ecto.Changeset<
   action: :insert,
   changes: %{},
   errors: [
     delegator_id: {"can't be blank", [validation: :required]},
     delegate_id: {"can't be blank", [validation: :required]},
     organization_id: {"can't be blank", [validation: :required]}
   ],
   data: #LiquidVoting.Delegations.Delegation<>,
   valid?: false
 >}
```

jinjagit:
All suggested changes addressed

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
oliverbarnes: I'm thinking of each version of the delegation as separate delegations, as it would be from a user's point of view, even if behind the scenes we're modifying the same record on the db. 

With a query using an old delegate as param, we'd be asserting that a query for the old delegation doesn't return anything, and it'd be more explicit than counting the delegations for an organization.

jinjagit: I kind of understand what you mean, but I'm confused by the fact that the only way to query a specific delegation (old or new) is by id (at absinthe level, for user). So, the only query possible to check if an 'old' delegation exists (for same delegate or not) is using the id, which will be used by the 'new' updated delegation (and thus it exists, even if it has a different delegate).

jinjagit: Maybe we could do it by querying delegations for an org, and returning delegate/delegator ids, etc. in the result, and asserting that a specific combo of delegation specific values does not exist after an update to a delegation?

oliverbarnes: I see what you mean - the resolver for querying a single delegation only accepts an id right now! I had Delegations.get_delegation!/3 in mind, not the resolver. In this case I think the way you're doing it now is the best solution, I don't think asserting the missing values would add much. I'm going to mark this as resolved and open an issue for updating the resolver, I think we need querying by delegator and delegate at that level as well.

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
oliverbarnes: 👍 I'd do that, the code will be tighter and less misleading. The way it is now it seems we expect a potential empty map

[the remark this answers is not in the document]
jinjagit: yup, it is removed +1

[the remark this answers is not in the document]
oliverbarnes: 🤦‍♂️ the code around the thread is stale (again). I'm proposing a review process change on slack.

thanks for addressing this, marking it as resolved!

test/liquid_voting/delegations_test.exs
oliverbarnes: Is this done?

test/liquid_voting/delegations_test.exs
oliverbarnes: I think this went unaddressed as well?

test/liquid_voting/delegations_test.exs
jinjagit: I think so. I created and used 
```
valid_proposal_specific_attrs: %{
          delegator_id: delegator.id,
          delegate_id: delegate.id,
          organization_id: organization_id,
          proposal_url: proposal_url
        },
        update_proposal_specific_attrs: %{
          delegator_id: delegator.id,
          delegate_id: another_delegate.id,
          organization_id: organization_id,
          proposal_url: proposal_url
        }
```
Which I assumed == using factories.

jinjagit: It's certainly a cleaner approach.

test/liquid_voting/delegations_test.exs
jinjagit: I added:
```
test "upsert_delegation/1 with valid proposal_specific delegation data creates a delegation",
         context do
      assert {:ok, %Delegation{} = delegation} =
               Delegations.upsert_delegation(context[:valid_proposal_specific_attrs])
end

test "upsert_delegation/1 with valid global delegation data creates a delegation", context do
      assert {:ok, %Delegation{} = delegation} =
               Delegations.upsert_delegation(context[:valid_attrs])
end
```

test/liquid_voting/delegations_test.exs
oliverbarnes: Ignore what I said earlier about the same delegation being created twice, I had mis-read the setup code.

What I had in mind was this:

```elixir
# factory-created original delegation and new delegate, explicitly not code under test
delegation = insert(:delegation)
new_delegate = insert(:participant)

args = {% 
  delegator_id: delegation.delegator_id,
  delegate_id: new_delegate.id,
  organization_id: valid_attrs[:organization_id]
}
assert {:ok, %Delegation{} = modified_delegation} = Delegations.upsert_delegation(args)
assert modified_delegation.id == delegation.id
assert modified_delegation.organization_id == delegation.organization_id
assert modified_delegation.delegate_id != delegation.delegate_id
assert modified_delegation.delegator_id == delegation.delegator_id

# no need to list delegations and count them, since we've already asserted the same record id
```

Your latest update did make things clearer, though. Just wondering whether setup is getting bloated. Perhaps we can leave it as is, and refactor the test file later.

Your call. Only hard request here is to replace listing delegations to assert count, and just assert on the same record id.

oliverbarnes (approved):
👍 just one last comment pending, but I'm going to approve and unblock this so Niklas can go ahead and give it another pass.

test/liquid_voting/delegations_test.exs
jinjagit: I can remove the count, but I didn't because  you said "I don't think we need to test the delegation count here, since we're already testing it on the LiquidVoting.Delegations layer." in a comment about the absinthe tests.

I don't think we need to test the delegation count here, since we're already testing it on the LiquidVoting.Delegations layer.

oliverbarnes: Sure, but asserting the id ensure the count is 1 as well. It's the same record and so the count didn't change

jinjagit: Got it. I'll remove it after my pairing.

oliverbarnes: Another thing to keep in mind is, it's a good practice for unit tests to exercise as little code external to actual code under test as possible.

Asserting the id is more reliable since it doesn't depend on another function. We're just asserting over the persisted record, which is an outcome of the upsert itself. If `Delegations.list_delegations` gets a bug, it won't affect it.

Same reasoning applies to using factories instead of using `Delegations.create_delegation`, because the factory just populates the db. Currently if `Delegations.create_delegation` gets a bug down the line, this test for upserts will still run, giving us a false negative (or positive, depending on how one sees it). A false breakage detection: upsert actually works, but its test breaks because we're coupling it with an external function to the upsert.

So I guess I am requesting for `Delegations.create_delegation` to be replaced with a factory, sorry man =D

oliverbarnes (changes requested):
Last change requested on: https://github.com/liquidvotingio/api/pull/112#discussion_r477443174

test/liquid_voting/delegations_test.exs
oliverbarnes: (no rush: async, etc)

niklaslong: +1 on leveraging `ex_machina` in this case as long as we have integration tests covering the _units_ together.

jinjagit: No problem. I think I initially misunderstood what you wanted. I _think_ I get it now.

jinjagit: How do add a `proposal_url` field + value to `insert(:delegation)` factory item?

oliverbarnes: > No problem. I think I initially misunderstood what you wanted. I think I get it now.

and I probably was a little convoluted in my explaining it 😄 this discussion on tests on Friday will be good. And in general we should pair more to avoid these long running reviews, which can be taxing

jinjagit: done: removed delegations counts + use factories to replace calls to create_delegations/1

test/liquid_voting/delegations_test.exs
```diff
@@ -10,22 +10,36 @@ defmodule LiquidVoting.DelegationsTest do
...
           delegate_id: nil,
           organization_id: nil
+        },
+        valid_proposal_specific_attrs: %{
+          delegator_id: delegator.id,
+          delegate_id: delegate.id,
+          organization_id: organization_id,
+          proposal_url: proposal_url
+        },
+        update_proposal_specific_attrs: %{
```
oliverbarnes, line 38: I don't see this used anywhere, looks like it's orphaned by latest changes

oliverbarnes (changes requested):
Almost there! One last change, and @niklaslong can give it another pass and merge

test/liquid_voting/delegations_test.exs
jinjagit: Done - good eyes!

oliverbarnes (approved):

oliverbarnes:
All yours, @niklaslong :)

niklaslong (approved):
Great work @jinjagit!

Issue #115, opened 2020-08-27
Add tests ensuring more double delegations scenarios don't happen


oliverbarnes:
Currently we have a test against duplicate global delegations:

```elixir
setup do
...
      valid_attrs: %{
          delegator_id: delegator.id,
          delegate_id: delegate.id,
          organization_id: Ecto.UUID.generate()
        }
...
test "create_delegation/1 with duplicate data returns error changeset", context do
  Delegations.create_delegation(context[:valid_attrs])
  assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(context[:valid_attrs])
end
```

We also need:

- 2 delegations by same delegator to same delegate on same proposal
- 2 delegations by same delegator to same delegate, 1 global and 1 proposal-specific

We're expecting these to all pass with current over-restrictive unique index, but they might break when it's removed on https://github.com/liquidvotingio/api/issues/24, and when allowing multiple delegations to same delegate but to different proposals (not sure we have an issue for this, actually? 🤔)

oliverbarnes:
Added this to the top of the Kanban board.

PR #116, opened 2020-08-28
Add tests of double delegations scenarios

jinjagit:
closes [#115](https://github.com/liquidvotingio/api/issues/115)

I added 3 tests:

1. `create_delegation/1 with duplicate proposal-specific data returns error changeset`  
2. `create_delegation/1 with proposal-specifc data returns error if global delegation for same delegator/delegate pair exists`
3. `create_delegation/1 with global delegation data returns error if proposal-specific delegation for same delegator/delegate pair exists`

### DIFF
--- a/test/liquid_voting/delegations_test.exs
+++ b/test/liquid_voting/delegations_test.exs
@@ -14,6 +14,7 @@ defmodule LiquidVoting.DelegationsTest do
       proposal_url = "https://www.someorg/proposalX"
 
       [
+        proposal_url: proposal_url,
         valid_attrs: %{
           delegator_id: delegator.id,
           delegate_id: delegate.id,
@@ -69,7 +70,7 @@ defmodule LiquidVoting.DelegationsTest do
       assert {:ok, %Delegation{} = delegation} = Delegations.create_delegation(args)
     end
 
-    test "create_delegation/1 with duplicate data returns error changeset", context do
+    test "create_delegation/1 with duplicate data returns error changeset" do
       original_delegation = insert(:delegation)
 
       args = %{
@@ -83,7 +84,7 @@ defmodule LiquidVoting.DelegationsTest do
 
     test "create_delegation/1 with duplicate proposal-specific data returns error changeset",
          context do
-      original_delegation = insert(:delegation, proposal_url: "https://www.someorg/proposalX")
+      original_delegation = insert(:delegation, proposal_url: context[:proposal_url])
 
       args = %{
         delegator_id: original_delegation.delegator_id,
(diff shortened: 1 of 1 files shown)


jinjagit:
Would we prefer using a factory for the first `create_delegation/1` usage in each test, then building args (partially from that first factory delegation) for the second `create_delegation/1` usage (in the assertion)?

oliverbarnes:
Yes, I think so. And it's first time I see _context_ used like that, better to set it up earlier and keep the test cleaner

jinjagit:
I've used factory items in all the tests of duplicate delegations.

I could open an issue to change all tests to that format (where possible)... but probably best to wait for our upcoming discussion of testing.

test/liquid_voting/delegations_test.exs
```diff
@@ -38,6 +38,8 @@ defmodule LiquidVoting.DelegationsTest do
       ]
     end
 
+    @proposal_url "https://www.someorg/proposalX"
```
oliverbarnes, line 41: Usually we either have setup contexts or module attributes like this, but not both at the same time on the same test file. So it's clear where all the test data is. I'd just add this to the setup context in this case

oliverbarnes:
Yeah, and they can also be changed incrementally when code surrounding it is touched (boy scout rule)

oliverbarnes (changes requested):
Looking good, just that one comment

test/liquid_voting/delegations_test.exs
jinjagit: Do you mean I should change line 14 to: `proposal_url = @proposal_url` and leave line 41 as is? Or something else?

I need a way to sett the proposal url in lines like line 88: `original_delegation = insert(:delegation, proposal_url: @proposal_url)` Where I am trying to avoid specifying the same url string in multiple tests (though I don't know why that matters really)

jinjagit: So, I removed the attribute `@proposal_url =  "https://www.someorg/proposalX"` and edited the 3 occurrences of its use in tests to: `original_delegation = insert(:delegation, proposal_url: "https://www.someorg/proposalX")`.  
This doesn't feel like what you meant, though.

oliverbarnes: I mean adding it to the setup so you can access it with `context[:proposal_url]` :)

jinjagit: Done. Thanks for clarifying.

oliverbarnes (approved):
💪 thanks man!

Issue #45, opened 2020-06-01
Validate delegate and delegator are different


oliverbarnes:
Right now one can delegate to themselves. It doesn't actually change the weight calculation (so it can't be exploited to increase one's own weight), but we should validate against it to signal a mistake by the api consumer.

PR #117, opened 2020-08-31
Validate delegator and delegate are different

jinjagit:
closes [#45](https://github.com/liquidvotingio/api/issues/45)  

Q: Do we want the changeset error to apply to both fields (`delegator_id` & `delegate_id`)? I have only applied it to `delegate_id` (arbitrary choice).  
Q: Do we want a different error wording? I have suggested `delegate_id: ["delegator and delegate must be different"]`

### DIFF
--- a/test/liquid_voting/delegations_test.exs
+++ b/test/liquid_voting/delegations_test.exs
@@ -60,6 +60,19 @@ defmodule LiquidVoting.DelegationsTest do
       assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(context[:invalid_attrs])
     end
 
+    test "create_delegation/1 with same participant as delegator and delegate returns error changeset",
+         context do
+      participant = insert(:participant)
+
+      args = %{
+        delegator_id: participant.id,
+        delegate_id: participant.id,
+        organization_id: Ecto.UUID.generate()
+      }
+
+      assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(args)
+    end
+
     test "create_delegation/1 with proposal url creates a delegation", context do
       # Test long urls while at it
       proposal_url = """
--- a/test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
+++ b/test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
@@ -191,11 +191,34 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegationTest do
       assert details == %{delegate_id: ["can't be blank"]}
     end
 
-    test "with identical emails for delegator and delegate" do
+    test "with identical emails for delegator and delegate", context do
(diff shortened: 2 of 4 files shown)


niklaslong:
In response to your outstanding questions, I think applying the error to the `delegate_id` makes sense as that's the user choice which results in the error. I also like the current wording.

test/liquid_voting/delegations_test.exs
```diff
@@ -60,6 +60,19 @@ defmodule LiquidVoting.DelegationsTest do
...
+         context do
+      participant = insert(:participant)
+
+      args = %{
+        delegator_id: participant.id,
+        delegate_id: participant.id,
+        organization_id: Ecto.UUID.generate()
+      }
+
+      assert {:error, %Ecto.Changeset{}} = Delegations.create_delegation(args)
```
niklaslong, line 73: I wonder if we should assert on `changeset.errors` here as well? I looked at the rest of the file and we don't seem to be doing it but thought I'd raise the question. Checking the error at the GraphQL level may be enough.

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
```diff
@@ -190,6 +190,29 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegationTest do
       assert message == "Could not create delegation"
       assert details == %{delegate_id: ["can't be blank"]}
     end
+
+    test "with identical emails for delegator and delegate" do
```
niklaslong, line 194: We may need a test for creating a delegation with IDs as well?

niklaslong (commented):
Great work! 🎉 
I left a few comments but this is mostly ready to go.

test/liquid_voting/delegations_test.exs
jinjagit: Yeah, we seem to have a convention of only checking the exact error (rather than just checking for any error) at the absinthe layer only. An issue for the upcoming test discussion?

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
jinjagit: Good point. I'll add this.

test/liquid_voting/delegations_test.exs
niklaslong: Yep, sounds good!

test/liquid_voting_web/absinthe/mutations/create_delegation_test.exs
jinjagit: done

niklaslong (approved):
LGTM, all yours @oliverbarnes

oliverbarnes (approved):
Looks 💯

PR #120, opened 2020-09-02
Correct error for missing delegate field

### LINKED ISSUE #118: Correct error reporting for missing fields when creating delegation
Currently, `create_delegation/1` with attributes that are missing a single email field (`delegator-email` or `delegate_email`) returns an error changeset which reports both `delegate_id` and `delegator_id` were blank.  

This can also be seen in the mismatch between the parameters passed into the `createDelegation` mutation and the assertions (which currently pass) in this test:  
```
test "with missing fields" do
      query = """
      mutation {
        createDelegation(delegatorEmail: "#{@delegator_email}") {
          delegator {
            email
            name
          }
          delegate {
            email
            name
          }
        }
      }
      """
      {:ok, %{errors: [%{message: message, details: details}]}} =
        Absinthe.run(query, Schema, context: %{organization_id: Ecto.UUID.generate()})
      assert message == "Could not create delegation"
      assert details == %{delegate_id: ["can't be blank"], delegator_id: ["can't be blank"]}
    end
```
So, this test also needs correcting when the fix is implemented.

jinjagit:
closes #118  

**TLDR**: This 'fixes' the problem, but I don't particularly like the fix, and it perhaps raises more questions than it answers. This PR aims to stimulate discussion on the topic.  

**The issue:**  
createDelegation mutation with only 1 email field provided, misleadingly leads to an error of: `Could not create delegation ... %{delegate_id: ["can't be blank"], delegator_id: ["can't be blank"]}` This error does not help the user identify the missing email field.  

**The workaround:**     
Detect such a missing field (and the presence of one email field), and add the complimentary email field with a value of nil. Now a missing `:delegator_email` returns an error of `Could not create delegate ... email: ["can't be blank"]`, which is more useful.  

**Observations:**  
I have placed the 'fix' in the beginning of the `lib/liquid_voting_web/resolvers/delegations.ex create_delegation/3` function. It could go elsewhere (pattern matching in `lib/liquid_voting/delegations.ex create_delegation/1`, for example).  

**Questions raised:**  
This seems to be a product of having 2 different ways to identify participants at the absinthe layer (ids & emails). Should we be handling all possible cases? E.g. 1 id & 1 email, 2 ids, 2 emails, only 1 id, only 1 email, or nothing. We might well now be handling all these cases appropriately, but I would have to test more thoroughly to check, and I wanted to know if this issue signals a need for a more fundamental shift in approach, or just a different approach, before I start down that road.   
Also, does none of this matter, since the user UI will 'decide' how to fill in fields and these edge cases will never arise (is this YAGNI)?

Thoughts?

### DIFF
--- a/lib/liquid_voting_web/resolvers/delegations.ex
+++ b/lib/liquid_voting_web/resolvers/delegations.ex
@@ -13,34 +13,70 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
   # with create_delegation_with_valid_arguments/1
   def create_delegation(_, args, %{context: %{organization_id: organization_id}}) do
     args
+    |> validate_participants
     |> case do
-      %{delegator_email: _, delegate_email: _} = args ->
+      {:ok, args} ->
         args
+        |> Map.put(:organization_id, organization_id)
+        |> Delegations.create_delegation()
+        |> case do
+          {:ok, delegation} ->
+            {:ok, delegation}
 
-      # if delegator_email field exists, add delegate_email field with nil value
-      %{delegator_email: _} = args ->
-        Map.put(args, :delegate_email, nil)
+          {:error, changeset} ->
+            {:error,
+             message: "Could not create delegation",
+             details: ChangesetErrors.error_details(changeset)}
 
-      # if delegate_email field exists, add delegator_email field with nil value
-      %{delegate_email: _} = args ->
-        Map.put(args, :delegator_email, nil)
+          {:error, name, changeset, _} ->
+            {:error,
+             message: "Could not create #{name}",
+             details: ChangesetErrors.error_details(changeset)}
+        end
 
-      _ ->
-        args
+      {:error, message} ->
+        {:error, message}
     end
-    |> Map.put(:organization_id, organization_id)
-    |> Delegations.create_delegation()
+  end
+
+  defp validate_participants(args) do
+    args
     |> case do
-      {:ok, delegation} ->
-        {:ok, delegation}
+      %{delegator_email: _, delegate_email: _} ->
+        {:ok, args}
+      
+      %{delegator_id: _, delegate_id: _} ->
+        {:ok, args}
 
-      {:error, changeset} ->
-        {:error,
-         message: "Could not create delegation", details: ChangesetErrors.error_details(changeset)}
+      # if delegator_email field exists, but no delegate_email field exists
+      %{delegator_email: _} ->
+        {:error, %{
+          message: "Could not create delegation",
+          details: %{delegate_email: ["field not found"]}
+        }}
 
-      {:error, name, changeset, _} ->
-        {:error,
-         message: "Could not create #{name}", details: ChangesetErrors.error_details(changeset)}
+      # if delegate_email field exists, but no delegator_email field exists
+      %{delegate_email: _} ->
+        {:error, %{
+          message: "Could not create delegation",
+          details: %{delegator_email: ["field not found"]}
+        }}
+
+      # if delegator_id field exists, but no delegate_id field exists
+      %{delegator_id: _} ->
+        {:error, %{
+          message: "Could not create delegation",
+          details: %{delegate_id: ["field not found"]}
+        }}
+
+      # if delegate_id field exists, but no delegator_id field exists
+      %{delegate_id: _} ->
+        {:error, %{
+          message: "Could not create delegation",
+          details: %{delegator_id: ["field not found"]}
+        }}
+
+      _ -> {:error, "some generic error - as yet undecided"}
     end
   end
 
--- a/lib/liquid_voting_web/resolvers/delegations.ex
+++ b/lib/liquid_voting_web/resolvers/delegations.ex
@@ -13,7 +13,7 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
   # with create_delegation_with_valid_arguments/1
   def create_delegation(_, args, %{context: %{organization_id: organization_id}}) do
     args
-    |> validate_participant_fields_provided
+    |> validate_participant_args
     |> case do
       {:ok, args} ->
         args
@@ -39,7 +39,7 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
     end
   end
 
-  defp validate_participant_fields_provided(args) do
+  defp validate_participant_args(args) do
     args
     |> case do
       %{delegator_email: _, delegate_email: _} ->
--- a/lib/liquid_voting_web/resolvers/delegations.ex
+++ b/lib/liquid_voting_web/resolvers/delegations.ex
@@ -48,41 +48,34 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
       %{delegator_id: _, delegate_id: _} ->
         {:ok, args}
 
-      # if delegator_email field exists, but no delegate_email field exists
+      # delegator_email field provided, but no delegate_email field provided
       %{delegator_email: _} ->
-        {:error,
-         %{
-           message: "Could not create delegation",
-           details: %{delegate_email: ["field not found"]}
-         }}
+        field_not_found_error(%{delegate_email: ["field not found"]})
 
-      # if delegate_email field exists, but no delegator_email field exists
+      # delegate_email field provided, but no delegator_email field provided
       %{delegate_email: _} ->
-        {:error,
-         %{
-           message: "Could not create delegation",
-           details: %{delegator_email: ["field not found"]}
-         }}
+        field_not_found_error(%{delegator_email: ["field not found"]})
 
-      # if delegator_id field exists, but no delegate_id field exists
+      # delegator_id field provided, but no delegate_id field provided
       %{delegator_id: _} ->
-        {:error,
-         %{
-           message: "Could not create delegation",
-           details: %{delegate_id: ["field not found"]}
-         }}
+        field_not_found_error(%{delegate_id: ["field not found"]})
 
-      # if delegate_id field exists, but no delegator_id field exists
+      # delegate_id field provided, but no delegator_id field provided
       %{delegate_id: _} ->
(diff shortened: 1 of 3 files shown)


oliverbarnes:
> Should we be handling all possible cases? E.g. 1 id & 1 email, 2 ids, 2 emails, only 1 id, only 1 email, or nothing.

Just two cases: either both ids are present, or both emails. So if a platform chooses to use the id strategy, they need to provide both ids from existing participants. If they want to use emails instead, either from existing participants or new ones, they need to provide them for both participants as well.

And I see two ways to handle this:

One might be to add a resolver-level validator for the presence of either the two ids or the two emails, and if args don't comply then `Delegations.create_delegation()` never gets called and a custom error message is returned.

The other might be to let `Delegations.create_delegation()` execute and then check the error changeset for the missing id errors. If found, then return a custom error message for the resolver.

What do you think?

jinjagit:
> One might be to add a resolver-level validator for the presence of either the two ids or the two emails, and if args don't comply then `Delegations.create_delegation()` never gets called and a custom error message is returned.
> 
> The other might be to let `Delegations.create_delegation()` execute and then check the error changeset for the missing id errors. If found, then return a custom error message for the resolver.

I like the former, as it would enable helpful errors and also is more efficient (since it does not hit the create_delegation functions if an error is raised beforehand). The latter would need work to avoid the misleading errors that come from submitting one or zero emails (since these go to the id-based create_delegation/1 and return 'id can't be blank' for both delegator and delegate (regardless of whether one email was provided or not).

davefrey:
I like three former as well. I'd extract that case though so the pipeline reads in a meaningful way, something like `|> Delegations.validate_participants`

davefrey:
Hmm, now I see `Delegate.validate_participants_different(changeset)` which is a good sibling.
I'm too new to this code and elixir, but couldn't we have a 
`|> Delegate.validate_participant_identities(changeset)` and be specific in the error? `Multi` would rollback the `Voting.upsert_participant` calls right?

oliverbarnes:
It would, but that happens in the data context level, which the former solution tries to avoid :)

I like `validate_participants`, or 
`validate_participant_args`

I'm not sure about which is more efficient, actually. Ecto and postgres are both fast, but so is pattern matching. Difference might be negligible, would need to benchmark. Checking the id error changeset is probably not a lot of work either, actually. 

But separation of concerns and clarity of intent are good reasons to do it at the resolver level, imho

jinjagit:
Cool, I'll write a validate_participants function at the resolver level, with suitable error messages if not receiving 2 * ids or 2 * emails. Thanks for the input!  

Also switching this to draft / WIP for now.

jinjagit:
First attempt at validation at the resolver level.  
I decided to try writing comments in doc format. Not sure how useful the examples are. Thoughts?

oliverbarnes (approved):
Looks great 👌 Nothing to add!

@niklaslong would you like to have a look?

lib/liquid_voting_web/resolvers/delegations.ex
```diff
@@ -8,27 +8,109 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
...
+      details: %{delegate_email: ["can't be blank"]},
+      message: "Could not create delegation"
+    }}
+  """
   def create_delegation(_, args, %{context: %{organization_id: organization_id}}) do
     args
-    |> Map.put(:organization_id, organization_id)
-    |> Delegations.create_delegation()
+    |> validate_participant_args
     |> case do
```
niklaslong, line 50: Suggestion: refactor this to use with, something like this: 

```elixir
  with {:ok, args} <- validate_participant_args(args),
     {:ok, delegation} <- Delegations.create_delegation(args) do
  {:ok, delegation}
else
  error -> error
end
```

(`Map.put(args, :organisation_id)` could be done beforehand)

lib/liquid_voting_web/resolvers/delegations.ex
```diff
@@ -8,27 +8,109 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
...
-        {:error,
-         message: "Could not create #{name}", details: ChangesetErrors.error_details(changeset)}
+      {:error, message} ->
+        {:error, message}
     end
   end
 
+  defp validate_participant_args(args) do
+    args
+    |> case do
```
niklaslong, line 77: I'd remove the pipe here (as there is only one)

niklaslong (commented):
Great work, I left some suggestions!

lib/liquid_voting_web/resolvers/delegations.ex
jinjagit: I like this idea, but how do I include the 3 different types of error tuples that could be returned (1 type from the call to `validate_participant()` and 2 different types possible from call to `create_delegation()`) in the `with` structure you suggest?  
There seems to be only 1 error case, in the `else` block, whilst we could errors for either call.

niklaslong: You could move the three errors associated with `create_delegation` to the else block and bubble up the ones from `validate_participant` with the catch-all.

jinjagit: Do you mean this kind of approach?  
```
def convert_to_csv(source_filename) do
    with {:l1, {:ok, :xlsx}} <- {:l1, detect_file_type(source_filename)},
         {:l2, {:ok, csv_filename}} <- {:l2, convert_excel_to_csv(source_filename)} do
  {:ok, csv_filename}
else
  {:l1, {:error, error}} -> {:error, error}  # (such as file not found)
  {:l1, {:ok, _other}} -> {:error, :not_an_excel_file}
  {:l2, error} -> error
end
```

niklaslong: Something else I've noticed: we have two different forms for the error messages. Might be nice to homogenise this (later PR of course, but worth discussing, I think)?

niklaslong: > Do you mean this kind of approach?

Essentially you just need to move those errors for `create_delegation` somewhere, and I think the else block is fine as we don't want to introduce another wrapper around that function.

jinjagit: Agree. Errors need some love. Maybe create an issue?

jinjagit:
@niklaslong I have implemented your suggested changes. Thanks!

lib/liquid_voting_web/resolvers/delegations.ex
niklaslong: Ah my bad, you don't need to wrap the structs in steps. I was thinking: 

```elixir
 with {:ok, args} <- validate_participant_args(args),
     {:ok, delegation} <- Delegations.create_delegation(args) do
  {:ok, delegation}
else
  {:error, changeset} -> handle()
  {:error, name, changeset, _} -> handle()
  error -> error
end
```

Though I now realise that for this to work, you'd need to change the _shape_  of the errors in the validation function to `{:error, message, detail}`
```

Comment (jinjagit) [reply lib/liquid_voting_web/resolvers/delegations.ex]: OK, I can try that.  

BTW, I based the error structure on `assert details == %{email: ["is invalid"]}` from `test/liquid_voting_web/absinthe/mutations/create_participant_test.exs`

Comment (niklaslong) [reply lib/liquid_voting_web/resolvers/delegations.ex]: I see, could you add some detail to the error situation in #123?

Comment (jinjagit) [reply lib/liquid_voting_web/resolvers/delegations.ex]: To be clear, what are you seeing as the default / desired error structure?  Is it this:  
`{:ok, %{errors: [%{message: message, details: details}]}}`

Comment (niklaslong) [reply lib/liquid_voting_web/resolvers/delegations.ex]: I don't know what the default or desired structure is, that's the reason for opening #123. However, the structure we've had in the resolvers up until now is `{:error, message, detail}`.

Comment (jinjagit) [reply lib/liquid_voting_web/resolvers/delegations.ex]: I'm disappearing down a bit of a rabbit hole here, trying to get the errors to play well with the error -> error part, since create_delegation() can also return errors from upsert_participant, and these errors are not all formatted the way you are envisioning. Thus, I would need to change the error shapes elsewhere (as in the example above, which seems to be changeset error), which seems like scope creep.   
How about we go with the steps wrappers for now, and simplify after the errors issue is resolved?

Comment (niklaslong) [reply lib/liquid_voting_web/resolvers/delegations.ex]: In that case I'd revert to the nested case statements for now, we don't want the added complexity of steps in our return values. Let's save the rabbit-hole for another day once #123 is discussed.

Comment (jinjagit) [reply lib/liquid_voting_web/resolvers/delegations.ex]: Actually, this almost works...
```
with {:ok, args} <- validate_participant_args(args),
         {:ok, delegation} <- Delegations.create_delegation(args) do
      {:ok, delegation}
    else
      {:error, message} ->
        {:error, message}

      {:error, changeset} ->
        {:error,
         message: "Could not create delegation", details: ChangesetErrors.error_details(changeset)}

      {:error, name, changeset, _} ->
        {:error,
         message: "Could not create #{name}", details: ChangesetErrors.error_details(changeset)}
    end
```
Just getting one (other) test throwing an error now.

jinjagit: @niklaslong I think I finally have a simpler version, in line with your ideas, but still deferring reworking the error formats until another issue. Thanks for helping me with this, and especially with understand the matching issue!

jinjagit:
@niklaslong I think I finally have a simpler version, in line with your ideas, but still deferring reworking the error formats until another issue. Thanks for helping me with this, and especially with understand the matching issue!

niklaslong (approved):
LGTM, thanks for the great work @jinjagit! 🎉

oliverbarnes (approved):
LGTM 👍 

And I'll take a look at the new issue on error message formatting soon

davefrey (approved):

PR #122, opened 2020-09-03
First step reorganization of create delegation tests

davefrey:
Still much to do for `describe` and `test` comments


oliverbarnes (approved):

oliverbarnes (approved):
Fixed conflicts (my bad on not reviewing earlier), all green now. Shipping it

Issue #80, opened 2020-07-16
A global deleteDelegation mutation deletes proposal-specific delegation


oliverbarnes:
When sending a `deleteDelegation` mutation without a `proposalUrl` field, it finds a delegation with a proposal_url set and deletes it.

The expected behavior should be to find a global delegation, and ignore proposal-specific delegations.

oliverbarnes:
Fixed on https://github.com/liquidvotingio/api/pull/132

Issue #125, opened 2020-09-10
Resolve conflict between global and proposal-specific delegations to same delegate


jinjagit:
Following #109, it is possible to create both a global and proposal-specific delegation for the same delegator/delegate pair, but this is not the desired behaviour.  

Specifically we wish to treat 2 separate cases:  

1. If a proposal-specific delegation creation is attempted, when an existing global delegation for same delegator/delegate pair exists, the new delegation creation should be rejected.  
2. If a global delegation is created, then any existing proposal-specific delegations for same delegator/delegate pair should be deleted.

oliverbarnes:
👌 well described

jinjagit:
Note to self: There are 3 skipped tests in `test/liquid_voting/delegations_test.exs` that need addressing in this issue + probably need to create some absinthe tests to go with this issue and the cases it covers.

oliverbarnes:
Trying to summarize delegation constraints, for sanity :sweat_smile:. 

**At any given moment a given participant can only have, as a delegator:**
- 1 global delegation
- multiple proposal-specific delegations, as long as:
  - they're to delegates other than the global delegate
  - and each delegation has a different proposal

**At no point can a delegator have:**
- two global delegations
- a global delegation and a proposal-specifc delegation to the same delegate
- two proposal-specific delegations with the same proposal url

jinjagit:
Closing this, as covered by #132 and #133, except for issue raised in #137 (which is now in progress)

PR #132, opened 2020-09-16
Avoid incorrect delegation deletion

jinjagit:
I'm having a mental blank here. I have created a test, but I have used a graphQL mutation to create a delegation before I then delete it. Please see test/liquid_voting_web/absinthe/mutations/delete_delegation_test.exs   
I would like to insert the delegation without using a graphQL mutation, but can't think how to do it. The created mutation needs to be proposal specific and I need to be able to use the field values from the created delegation in the delete delegation mutation. How can I do this? Thanks!

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -90,7 +90,6 @@ defmodule LiquidVoting.Delegations do
     delegator = Voting.get_participant_by_email!(delegator_email, organization_id)
     delegate = Voting.get_participant_by_email!(delegate_email, organization_id)
 
-    delegation =
       Delegation
       |> where(
         [d],
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -90,13 +90,6 @@ defmodule LiquidVoting.Delegations do
     delegator = Voting.get_participant_by_email!(delegator_email, organization_id)
     delegate = Voting.get_participant_by_email!(delegate_email, organization_id)
 
-    # Repo.get_by!(
-    #   Delegation,
-    #   delegator_id: delegator.id,
-    #   delegate_id: delegate.id,
-    #   organization_id: organization_id
-    # )
-
     delegation =
       Delegation
       |> where(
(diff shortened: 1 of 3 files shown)


jinjagit:
And the test is not even failing as I think it should (and as the manual test of graphQL suggest it should) :confounded:

oliverbarnes:
> I have used a graphQL mutation to create a delegation before I then delete it. Please see test/liquid_voting_web/absinthe/mutations/delete_delegation_test.exs
> I would like to insert the delegation without using a graphQL mutation, but can't think how to do it

So far we've been setting up absinthe tests with another absinthe request, and after our test discussions I think it's fine. We can later refactor by abstracting setup functions, Niklas pointed out.

In this case, you can use the data returned in the response as params for the request under test.

Does this make sense?

jinjagit:
> > I have used a graphQL mutation to create a delegation before I then delete it. Please see test/liquid_voting_web/absinthe/mutations/delete_delegation_test.exs
> > I would like to insert the delegation without using a graphQL mutation, but can't think how to do it
> 
> So far we've been setting up absinthe tests with another absinthe request, and after our test discussions I think it's fine. We can later refactor by abstracting setup functions, Niklas pointed out.
> 
> In this case, you can use the data returned in the response as params for the request under test.
> 
> Does this make sense?

Yup. I'll try again tomorrow, with that approach. Thanks.

jinjagit:
Current status (WIP): Appears to be working + tests passing. Needs manual testing + tests refactoring/restructuring (to improve clarity)

jinjagit:
I have created tests of `get_delegation!/3` because this PR effectively rests on the changes made to the global delegation clause of that function. These tests should probably be separated into an email-focused (vs. id-focused) testing file, but this seems beyong the scope of this PR.

oliverbarnes (approved):
Looks great, man 👏 

Only point needing further discussion is: we seem to be going back and forth between using factories and no in absinthe tests. We can debate that on slack :) I'll merge this right away to unblock #125

jinjagit:
> Looks great, man clap
> 
> Only point needing further discussion is: we seem to be going back and forth between using factories and no in absinthe tests. We can debate that on slack :) I'll merge this right away to unblock #125

I agree. We have a number of inconsistencies in tests (also in error formatting). I want to get around to working on both, but also want to get delegations working well first. Agree that group discussion is way to decide on what patterns to consistently use (and to create respective issues for) :+1:

oliverbarnes:
Closed #80

PR #133, opened 2020-09-21
Resolve conflicts for delegations with same participants

jinjagit:
See issue #125.

### Current status (early days):  
* 2 * failing tests (1 for each case). Test assertions will be more specific when changes fully implemented.  
* Current plan is something like this pseudo-code to modify upsert_delegation/1:  
```
    # Delegations = all delegations for delegator
    #
    #  if trying to create proposal delegation
    #    search Delegations for any global delegation with SAME DELEGATE (should be one or none)    
    #      if find global delegation
    #        return error -> "global delegation already exists"
    #  if trying to create global delegation
    #    search Delegations for any proposal delegations with SAME DELEGATE
    #      if find proposal(s) delegations
    #       delete all proposal delegations found
    #
    #  next (keep the existing functionality):
    #    if trying to create global delegation
    #      search Delegations for ANY GLOBAL delegation
    #        if found (should only be one or none)
    #          update
    #        else
    #          create new delegation 
    #    if trying to create proposal delegation
    #      search Delegations for SAME PROPOSAL delegation
    #        if found (should only be one or none)
    #          update
    #        else
    #          create new delegation
```

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -172,25 +172,26 @@ defmodule LiquidVoting.Delegations do
   def upsert_delegation(%{delegator_id: delegator_id} = attrs) do
     proposal_url = Map.get(attrs, :proposal_url)
 
-    # delegations = all delegations for delegator
-    #   search delegations for any with SAME DELEGATE
-    #     if find global and trying to create proposal
-    #       return error -> "global delegation already exists"
-    #     if find proposal(s) and trying to create global
-    #       delete all proposals found & create global
+    # Delegations = all delegations for delegator
+    #
+    #  search Delegations for any with SAME DELEGATE
+    #    if find global and trying to create proposal
+    #      return error -> "global delegation already exists"
+    #    if find proposal(s) and trying to create global
+    #      create new global delegation & delete all proposals found
     #  next (keep the existing functionality):
-    #     if trying to create global
-    #       search delegations for ANY GLOBAL
-    #         if found (should only be one or none)
-    #           update
-    #         else
-    #           create new delegation 
-    #     if trying to create proposal
-    #       search delegations for SAME PROPOSAL
-    #         if found (should only be one or none)
-    #           update
-    #         else
-    #           create new delegation
+    #    if trying to create global
+    #      search Delegations for ANY GLOBAL
+    #        if found (should only be one or none)
+    #          update
+    #        else
+    #          create new delegation 
+    #    if trying to create proposal
+    #      search delegations for SAME PROPOSAL
+    #        if found (should only be one or none)
+    #          update
+    #        else
+    #          create new delegation
 
     Delegation
     |> where([d], d.delegator_id == ^delegator_id)
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -175,23 +175,23 @@ defmodule LiquidVoting.Delegations do
     # Delegations = all delegations for delegator
     #
     #  if trying to create proposal delegation
-    #    search Delegations for any proposal delegations with SAME DELEGATE
-    #      if find global in Delegations
+    #    search Delegations for any global delegation with SAME DELEGATE (should be one or none)    
+    #      if find global delegation
     #        return error -> "global delegation already exists"
-    #  if trying to create global
-    #    search Delegations for any global delegation with SAME DELEGATE (should be one or none)
-    #      if find proposal(s) in Delegations
+    #  if trying to create global delegation
+    #    search Delegations for any proposal delegations with SAME DELEGATE
+    #      if find proposal(s) delegations
     #       delete all proposal delegations found
     #
     #  next (keep the existing functionality):
     #    if trying to create global delegation
-    #      search Delegations for ANY GLOBAL
+    #      search Delegations for ANY GLOBAL dlegation
     #        if found (should only be one or none)
     #          update
     #        else
     #          create new delegation 
     #    if trying to create proposal delegation
-    #      search Delegations for SAME PROPOSAL
+    #      search Delegations for SAME PROPOSAL delegation
     #        if found (should only be one or none)
     #          update
     #        else
@@ -229,13 +229,14 @@ defmodule LiquidVoting.Delegations do
   defp resolve_conflicts(delegations, delegate_id, proposal_url) do
     delegations =
       case proposal_url do
-        #  if trying to create global
-        #    search Delegations for any global delegation with SAME DELEGATE (should be one or none)
-        #      if find proposal(s) in Delegations
+        #  if trying to create global delegation
+        #    search Delegations for any proposal delegations with SAME DELEGATE
+        #      if find proposal(s) delegations
         #       delete all proposal delegations found
-        nil ->
-          IO.puts("global case")
 
+        # case: attempting to create global delegation
+        nil ->
+          # find any proposal delegations with SAME DELEGATE
           same_delegate_proposal_delegations =
             Enum.filter(delegations, fn d ->
               d.proposal_url != nil and d.delegate_id == delegate_id
@@ -253,8 +254,8 @@ defmodule LiquidVoting.Delegations do
           end
 
         #  if trying to create proposal delegation
-        #    search Delegations for any proposal delegations with SAME DELEGATE
-        #      if find global in Delegations
+        #    search Delegations for any global delegation with SAME DELEGATE (should be one or none)    
+        #      if find global delegation
         #        return error -> "global delegation already exists"
         _proposal_url ->
           IO.puts("proposal case")
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -202,8 +202,8 @@ defmodule LiquidVoting.Delegations do
     |> Repo.all()
     |> resolve_conflicts(delegate_id, proposal_url)
     |> case do
-      {:ok, delegations} -> delegations
-      # Find delegations of same type as in attrs (global vs proposal) for same delegator
+      {:ok, delegations_of_delegator} -> delegations_of_delegator
+      # Search for delegations of same type as in attrs (global vs proposal) for same delegator
       |> Enum.filter(fn d ->
         d.delegator_id == delegator_id and d.proposal_url == proposal_url
       end)
@@ -220,13 +220,13 @@ defmodule LiquidVoting.Delegations do
     end
   end
 
-  defp resolve_conflicts(delegations, delegate_id, _proposal_url = nil) do
-    #  When attemting global delegation creation:
+  defp resolve_conflicts(delegations_of_delegator, delegate_id, _proposal_url = nil) do
+    #  When attempting global delegation creation:
     #    Search Delegations for any proposal delegations with SAME DELEGATE
     #    and delete all proposal delegations found
 
     # same_delegate_proposal_delegations =
-    delegations
+    delegations_of_delegator
     |> Stream.filter(fn d ->
       d.proposal_url != nil and d.delegate_id == delegate_id
     end)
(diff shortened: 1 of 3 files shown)


jinjagit:
closes #125

lib/liquid_voting/delegations.ex
```diff
@@ -169,28 +178,88 @@ defmodule LiquidVoting.Delegations do
...
+         %{
+           message: "Could not create delegation.",
+           details: "A global delegation for the same participants already exists."
+         }}
+    end
+  end
+
+  # Searches the list of delegations for the specified delegator for delegations
+  # of the same type (global or for the same specific proposal) as in the
+  # attributes passed into upsert_delegation/1.
```
oliverbarnes, line 247: Looks for a participant's existing delegation of matching type (global or proposal-specific)

oliverbarnes (changes requested):
Looking good! I've left some comments on a first pass, and will go through the tests next.

lib/liquid_voting/delegations.ex
jinjagit: How about: 'Looks for a _delegator's_ existing delegation of matching type (global or proposal-specific)'

lib/liquid_voting/delegations.ex
```diff
@@ -169,28 +178,88 @@ defmodule LiquidVoting.Delegations do
...
+  # Searches list of delegations for the specified delegator for any
+  # proposal-specific delegations to the same delegate in the attributes passed
+  # into upsert_delegation/1. and deletes any such delegations found.
+  #
+  # Clause 2: Matches an attempt to upsert a proposal-specific delegation.
+  #
+  # Searches list of delegations for the specified delegator for global delegation
+  # to the same delegate in the attributes passed into upsert_delegation/1 and
+  # returns an error if such a delegation is found.
+  defp resolve_conflicts(delegations_of_delegator, delegate_id, _proposal_url = nil) do
```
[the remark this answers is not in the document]
jinjagit: Maybe the name used doesn't make my meaning clear enough. I intend to convey the fact that this is a list of delegations of the delegator that was passed into `attrs` in create_delegation, where the same participant is also delegator. Perhaps it's enough to make this clear in the comments, though?

lib/liquid_voting/delegations.ex
oliverbarnes: That works too. I'm just trying to have less repetition of "delegat*" as it makes both comments and code very hard to reason about, because the words "delegator", "delegation" and "delegate" are so similar.

jinjagit: Agree. It's good to get feedback, as it's hard to see the comments/code with fresh eyes when I know what it does. Though, in a week, I won't remember what it does without help from comments / var names, etc. :wink:

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
oliverbarnes: I get it, but this function doesn't really know that - it just filters a bunch of delegations. The context where the function is used is the one that cares about which subset is being passed in, and there that should be made clear. 

Another point to consider is that creating/updating a delegation is always done for a given delegator, semantically that's a given. The user story is "a participant delegates their vote". If that's a smart default, everything else will be much easier to reason about.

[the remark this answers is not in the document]
jinjagit: > creating/updating a delegation is always done for a given delegator, semantically that's a given. The user story is "a participant delegates their vote". If that's a smart default, everything else will be much easier to reason about.

That sounds like a good wiki entry.

lib/liquid_voting/delegations.ex
oliverbarnes: Definitely, and I know it can feel like nitpicking when asking for changes to comments, but it really pays off down the line, or in a week like you said :)

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
oliverbarnes: Good idea. Just added it: https://github.com/liquidvotingio/api/wiki/Notes-on-inline-docs-wording

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
```diff
@@ -57,6 +58,90 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegation.ExistingDelegation
     end
   end
 
+  describe "create global delegation for delegator with existing proposal delegations to same delegate" do
+    test "deletes proposal specific delegations for same delegator/delegate pair" do
+      # First, create a proposal-specific delegation.
```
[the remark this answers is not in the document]
oliverbarnes: same for the new tests below

oliverbarnes (changes requested):
Unit tests almost perfect, absinthe tests need to switch to factories

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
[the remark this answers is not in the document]
jinjagit: OK. Can you point me at a good example of 'using factories' in our code, so that i can use that as a model for the re-write?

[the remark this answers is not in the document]
oliverbarnes: what you're already doing on unit tests: https://github.com/liquidvotingio/api/pull/133/files/d4f812ea8b3821522939c51c3ef91113e2cfd98e#diff-66baa972d0de9e38c029f635297dd20fR212

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
```diff
@@ -7,9 +7,10 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegation.ExistingDelegation
   @delegate_email "delegate@email.com"
   @another_delegate_email "another-delegate@email.com"
   @proposal_url "https://www.proposal.com/my"
+  @another_proposal_url "https://www.proposal.com/another"
   @organization_id Ecto.UUID.generate()
 
-  describe "with existing global delegation" do
+  describe "create global delegation for delegator with existing global delegation to different delegate" do
```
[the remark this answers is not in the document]
jinjagit: I used your new wording, which is clearer, combined with the early part of the describe statements (which explains what type of delegation creation is being attempted - global v proposal), as the expected behavior differs depending on these contexts. Now we have:   

- create global delegation when a global delegation for a different delegate already exists
- create global delegation when proposal delegations to same delegate already exist
- create proposal delegation when delegation for same proposal already exists
- create proposal delegation when global delegation to same delegate already exists

jinjagit:
All changes suggested by @oliverbarnes addressed.

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
```diff
@@ -1,19 +1,28 @@
...
+        delegator: global_delegation.delegator,
+        another_delegate: another_delegate
+      ]
+    end
+
+    test "overwrites existing global delegation", context do
       query = """
       mutation {
-        createDelegation(delegatorEmail: "#{@delegator_email}", delegateEmail: "#{@delegate_email}") {
+        createDelegation(delegatorEmail: "#{context[:delegator].email}", delegateEmail: "#{
```
oliverbarnes, line 23: Sorry, I meant replacing the test setup that's now done through absinthe requests, like this request to create the existing global delegation, with factory creation.

Test setup as in test data state preparation, and not [ExUnit's setup callback](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html)

oliverbarnes (changes requested):

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
jinjagit: OK, but I don't quite understand what that would look like. The docs you link don't have absinthe examples, and I can't find any online. I want to know what using the ExUnit callback for a test setup looks like.

oliverbarnes: ...just noticed this test actually only has one request :) My comment applies to the next test on line 70, for example.

In other words, let's not use ExUnit's setup, move the factories inline in each test, and replace setup _requests_ with factories.

I hope I'm making sense =D

jinjagit: I'm sure you are making sense, but I don't understand :disappointed:

jinjagit: But I probably would, given an example.

oliverbarnes: Update: we hashed it out on a slack call. Just need to remove setup blocks and inline factories, and then remove extra request assertions that are already tested at the unit leve.

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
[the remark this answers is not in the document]
oliverbarnes: 👍

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
jinjagit: @oliverbarnes I think I've got it now :blush:

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
```diff
@@ -22,48 +22,59 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegation.ExistingDelegation
...
-      assert original_delegation["delegate"]["email"] == @delegate_email
+      assert updated_delegation["delegator"]["email"] == global_delegation.delegator.email
+      assert updated_delegation["delegate"]["email"] == another_delegate.email
+      assert updated_delegation["id"] == global_delegation.id
+    end
+  end
+
+  describe "create global delegation when proposal delegation to same delegate already exists" do
+    test "returns the new global delegation" do
+      proposal_delegation_1 = insert(:delegation_for_proposal)
```
oliverbarnes, line 41: `proposal_delegation`, no need for the `_1`

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
```diff
@@ -75,17 +86,26 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.CreateDelegation.ExistingDelegation
...
-      {:ok, %{data: %{"createDelegation" => original_delegation}}} =
-        Absinthe.run(query, Schema, context: %{organization_id: @organization_id})
+      {:ok, %{data: %{"createDelegation" => updated_delegation}}} =
+        Absinthe.run(query, Schema,
+          context: %{organization_id: proposal_delegation.organization_id}
+        )
+
+      assert updated_delegation["delegator"]["email"] == proposal_delegation.delegator.email
+      assert updated_delegation["delegate"]["email"] == another_delegate.email
+      assert updated_delegation["proposalUrl"] == proposal_delegation.proposal_url
```
oliverbarnes, line 96: I'd also assert the delegation id is still the same, as this is something the API consumer might care about when handling the delegation data on the frontend

oliverbarnes (approved):
Looking beautiful. Just two small final requests for changes :)

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
jinjagit: done

test/liquid_voting_web/absinthe/mutations/create_delegation/existing_delegations_test.exs
jinjagit: done

jinjagit:
last 2 suggestions addressed :thumbsup:

PR #134, opened 2020-09-21
Linode deploy

oliverbarnes:
Updates the deployment and smoke test jobs on pushes to use the Linode cluster.

It removes the intermediate gcloud auth step, and just passes Linode's kubeconfig in through a Github secret.

After merging all new deploys will be made to Linode and not to Google. So if the api.liquidvoting.io domain isn't resolving to the new IP yet, we won't be able to see the changes. So merging needs to be coordinated with domain migration (see https://github.com/liquidvotingio/deployment/issues/10)

Part of https://github.com/liquidvotingio/deployment/issues/35

### DIFF
--- a/.github/workflows/on_pull_request.yml
+++ b/.github/workflows/on_pull_request.yml
@@ -16,12 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
-    - name: Setup Linode cli
-      uses: brendon1555/setup-linode-cli@master
-      with:
-        LINODE_CLI_TOKEN: ${{ secrets.LINODE_TOKEN }}
-    - run: linode-cli --help
     - name: Deploy
+      with:
+        KUBECONFIG: ${{ secrets.LINODE_KUBE_CONFIG }}
       run: kubectl get pods
 
   # tests:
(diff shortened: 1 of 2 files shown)


niklaslong (approved):
LGTM, though I have to trust this does what is intended 🤞

jinjagit (changes requested):
Also, looks good to me, though as for @niklaslong, I don't know enough about kubernetes clusters to be sure of the details. However it works (responds appropriately when queried via graphiQL app/curls, and it deploys fine (re-running CD workflow).  
  
I suspect some kind of auth-key issue (secrets setup / communication of secrets to new server setup?) with the smoke-tests, however, as the test file now fails at the first test (not getting a 200 OK) when run via GitHub workflows against the live api, but runs fine when tested locally against up-to-date master branch of api (using up-to-date deployment branch for smoke tests).  
  
So, I'm not directly suggesting a change to this PR, but highlighting that something connected to it appears to perhaps not be working as intended (which may involve making changes to this file, or may involve a change elsewhere).

oliverbarnes:
Thanks, guys. I tested the deploy with on_pull_request.yml and it worked, but have to wait for merge to check the smoke test works with the new IP as well. Let's wait on merge until I get the auth keys issue sorted

jinjagit (approved):

PR #141, opened 2020-10-01
Refactor multis

oliverbarnes:
Reading through https://elixirschool.com/blog/ecto-multi/, it's clear that what `Multi.run` does is run _operations_, so naming them after the operations themselves is clearer than naming after resources.

Tweaked some variable namings as well.

(PR doesn't need to be squashed, commits are good to be merged as they are)

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -185,8 +185,8 @@ defmodule LiquidVoting.Delegations do
     |> Repo.all()
     |> resolve_conflicts(delegate_id, proposal_url)
     |> case do
-      {:ok, delegations_of_delegator} ->
-        delegations_of_delegator
+      {:ok, delegations} ->
+        delegations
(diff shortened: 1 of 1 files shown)


davefrey:
In addition to checking this visually, I confirmed `mix test cover` shows the changed code is well-covered, and also introduced a bug that failed in the same way before and after the refactor.

davefrey (approved):

oliverbarnes:
Thanks! Would you elaborate about the bug - is there an issue opened for it already?

davefrey:
There's not a real bug, I just poked part of the `create_delegation`
implementation (I changed a return value on one of the clauses) so that
I could see tests break, and compared the breakage before and after the
refactor. It's kind of overkill but I love testing my assumptions.

The theoretical risk this could help with is a green test suite that
stays green but doesn't test well enough to detect a change in behavior;
I wasn't thorough, it was just a sanity check, and a more active PR
review.

Issue #52, opened 2020-06-12
Update voting results / vote weights after each delegation


oliverbarnes:
Right now results and weights only get updated on vote creation/deletion and delegation deletion for a specific proposal. There's an issue for recalculating on deletion of global delegations, but we also need to recalculate on delegation creations, both proposal-specific and global

niklaslong:
I think we should indiscriminately update past votes for now and in future work out whether we want a voting period (#65)?

oliverbarnes:
Yeah, makes sense. Voting period (or result freeze) is definitely a separate issue and can of worms

Issue #24, opened 2020-04-28
Conflicting global and specific delegations


oliverbarnes:
Offshoot of #9:

If a global delegation is given to a delegate, but then another one is given specifically for a proposal by the same delegator to a third person, the former should still work for any other proposal.

jinjagit:
And also in the opposite order. If proposal specific delegation created, then if delegator subsequently creates a global delegation, only the proposal specific delegation is counted for that proposal result (and otherwsie the global delegation continues).
This may be best resolved when calculating voting results.

jinjagit:
Manual testing reveals that global and proposal-specific delegations correctly weight the votes of delegates when the delegations are created before the related votes, but not when votes are created before the related delegations. The issue of delegation creation not updating voting weights/results correctly is addressed in issue #52.  
  
I have attempted to create 2 absinthe tests in PR #XXX [update when PR created/merged] that test each of the above scenarios, with the latter being skipped for now, as is failing due to reason described above.

jinjagit:
Covered in #144

PR #144, opened 2020-10-07
Add test of vote counts when participant has global and proposal delegations

jinjagit:
This addresses issue #24, though really it highlights problems related to issue #52.  
  
Manual testing reveals that global and proposal-specific delegations correctly weight the votes of delegates when the delegations are created before the related votes, but not when votes are created before the related delegations. The issue of delegation creation not updating voting weights/results correctly is addressed in issue #52.

I have attempted to create 2 absinthe tests that test each of the above scenarios (delegations before votes, and vice-versa), with the latter being skipped for now, as is failing due to reason described above. I would expect to create more specific tests of the stages involved in the processes covered here, when addressing #52.  

I can imagine that there may well be better ways to test the issue than the 2 tests I have created here, but they are a good starting point for further discussion. Thoughts?


jinjagit:
Removed skipped test and improved setup and comments for new test.

oliverbarnes (changes requested):
Looking good 👍 Just two small requests: 

- update the PR title to inform this is about test coverage, rather than adding functionality
- add a blank line between the first and second, and third and fourth steps for legibility

jinjagit:
- Changed title to indicate this PR adds a test.  
- Added newlines to test comments.

oliverbarnes (approved):

PR #146, opened 2020-10-08
Use merge_attributes in delegation factory

jinjagit:
We should probably do this for other factories too.

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -5,33 +5,41 @@ defmodule LiquidVoting.Factory do
   alias LiquidVoting.Delegations.Delegation
   alias LiquidVoting.VotingResults.Result
 
-  def participant_factory do
-    %Participant{
+  def participant_factory(attrs) do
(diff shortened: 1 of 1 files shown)


oliverbarnes (changes requested):
Yeah, we should do it for any factories building associations and sharing attributes, like `vote_factory`

Would you also capitalize the beginning of the commit sentence, please? I've edited the PR title. 👮‍♂️

jinjagit:
> Yeah, we should do it for any factories building associations and sharing attributes, like `vote_factory`
> 
In this PR?

oliverbarnes:
Yes please - I asked for a separate PR also to add this, as it's outside the scope of the other one

oliverbarnes (changes requested):
`delegation_factory` also has associations with shared attrs

jinjagit:
I have now added merge_attributes to all the factories.

jinjagit:
Hold off one second on the (re)review. need to run mix format!

oliverbarnes (approved):

PR #147, opened 2020-10-10
Update voting results and weights after create delegation

### LINKED ISSUE #52: Update voting results / vote weights after each delegation

jinjagit:
Closes #52.

### DIFF
--- a/lib/liquid_voting/voting.ex
+++ b/lib/liquid_voting/voting.ex
@@ -83,10 +83,10 @@ defmodule LiquidVoting.Voting do
   end
 
   def list_votes_by_participant(participant_id, organization_id) do
-  Vote
-  |> where(participant_id: ^participant_id, organization_id: ^organization_id)
-  |> Repo.all()
-  |> Repo.preload([:participant])
+    Vote
+    |> where(participant_id: ^participant_id, organization_id: ^organization_id)
+    |> Repo.all()
+    |> Repo.preload([:participant])
   end
 
   @doc """
--- a/lib/liquid_voting/voting.ex
+++ b/lib/liquid_voting/voting.ex
@@ -71,15 +71,22 @@ defmodule LiquidVoting.Voting do
 
   ## Examples
 
-      iex> list_votes("https://docs.google.com/document/d/someid", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      iex> list_votes_by_proposal("https://docs.google.com/document/d/someid", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
       [%Vote{}, ...]
 
   """
-  def list_votes(proposal_url, organization_id) do
+  def list_votes_by_proposal(proposal_url, organization_id) do
     Vote
     |> where(proposal_url: ^proposal_url, organization_id: ^organization_id)
     |> Repo.all()
     |> Repo.preload([:participant])
+  end
+
+  def list_votes_by_participant(participant_id, organization_id) do
+  Vote
+  |> where(participant_id: ^participant_id, organization_id: ^organization_id)
+  |> Repo.all()
+  |> Repo.preload([:participant])
   end
 
   @doc """
--- a/lib/liquid_voting/voting.ex
+++ b/lib/liquid_voting/voting.ex
@@ -82,6 +82,15 @@ defmodule LiquidVoting.Voting do
     |> Repo.preload([:participant])
   end
 
+  @doc """
+  Returns the list of votes for a participant id and organization id
+
+  ## Examples
+
+      iex> list_votes_by_participant("cc1e2ea3-317e-4f40-97b4-06db6e48cd05", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      [%Vote{}, ...]
+
+  """
   def list_votes_by_participant(participant_id, organization_id) do
     Vote
     |> where(participant_id: ^participant_id, organization_id: ^organization_id)
--- a/lib/liquid_voting_web/resolvers/delegations.ex
+++ b/lib/liquid_voting_web/resolvers/delegations.ex
@@ -1,5 +1,5 @@
 defmodule LiquidVotingWeb.Resolvers.Delegations do
-  alias LiquidVoting.{Delegations, VotingResults}
+  alias LiquidVoting.{Voting, Delegations, VotingResults}
   alias LiquidVotingWeb.Schema.ChangesetErrors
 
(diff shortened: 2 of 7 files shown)


lib/liquid_voting_web/resolvers/delegations.ex
```diff
@@ -49,6 +49,21 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
...
+      case proposal_url do
+        # Global delegation: We find all votes of the delegate and update related voting result(s).
+        nil ->
+          Voting.list_votes_by_participant(delegation.delegate_id, delegation.organization_id)
+          |> Enum.each(fn vote ->
+            VotingResults.publish_voting_result_change(vote.proposal_url, vote.organization_id)
+          end)
+
+        # Proposal delegation: We update the voting result for the given proposal_url.
+        _proposal_url ->
```
jinjagit, line 63: 2 main options when delegation created is proposal-specific:  
  
1. Simply call publish_voting_result_change/2, which will call calculate_result!/2. This will create a new voting result with 0 votes both 'in favor' and 'against', if no votes have been cast for this proposal. I have used this approach.  
2. Check if delegate _has voted_ on this proposal (we have a get_vote!/3 that would find any such vote), and only call publish_voting_result_change/2 if vote exists. Otherwise, do nothing. This will mean there is no related voting result created if no delegate vote exists (votingResult will return 'null' for this proposal).  
  
If 2 is preferable, it easy to implement. Thoughts?

lib/liquid_voting/voting.ex
```diff
@@ -71,17 +71,33 @@ defmodule LiquidVoting.Voting do
 
   ## Examples
 
-      iex> list_votes("https://docs.google.com/document/d/someid", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      iex> list_votes_by_proposal("https://docs.google.com/document/d/someid", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
       [%Vote{}, ...]
 
   """
```
jinjagit, line 77: I renamed 'list_votes' to 'list_votes_by_proposal' to differentiate it from the new 'list_votes_by_participant/2'.

test/liquid_voting_web/absinthe/mutations/create_delegation/after_creating_related_votes_test.exs
```diff
@@ -0,0 +1,183 @@
...
+  use LiquidVotingWeb.ConnCase
+  import LiquidVoting.Factory
+
+  alias LiquidVotingWeb.Schema.Schema
+
+  describe "create proposal-specific delegation" do
+    test "after creating related proposal vote by delegate" do
+      # Insert vote and get related participant and proposal_url.
+      vote = insert(:vote, yes: true)
+      participant = vote.participant
```
oliverbarnes, line 11: Perhaps calling the participant a `voter` instead might be clearer?

test/liquid_voting_web/absinthe/mutations/create_delegation/after_creating_related_votes_test.exs
```diff
@@ -0,0 +1,183 @@
...
+          }
+        }
+      }
+      """
+
+      {:ok, %{data: %{"createDelegation" => delegation}}} =
+        Absinthe.run(query, Schema, context: %{organization_id: vote.organization_id})
+
+      assert delegation["proposalUrl"] == proposal_url
+      assert delegation["votingResult"]["inFavor"] == 2
```
oliverbarnes, line 37: Maybe add a comment here as well, explaining the assertion that the vote now carries a weight of 2 due to the delegation given to the voter?

test/liquid_voting_web/absinthe/mutations/create_delegation/after_creating_related_votes_test.exs
```diff
@@ -0,0 +1,183 @@
...
+      assert delegation["votingResult"]["against"] == 0
+    end
+  end
+
+  describe "create global delegation" do
+    test "after creating vote by delegate" do
+      # Insert vote and get related participant and proposal_url.
+      vote = insert(:vote, yes: false)
+      participant = vote.participant
```
oliverbarnes, line 46: voter here as well, if you're good with this change

test/liquid_voting_web/absinthe/mutations/create_delegation/after_creating_related_votes_test.exs
```diff
@@ -0,0 +1,183 @@
...
+  end
+
+  describe "create proposal-specific and global delegations" do
+    test "after creating related votes" do
+      # Insert vote_A and get related participant and proposal_url.
+      vote_A = insert(:vote, yes: true)
+      participant_A = vote_A.participant
+      proposal_A_url = vote_A.proposal_url
+
+      # Insert vote_B and get related participant asnd proposal_url.
```
oliverbarnes, line 106:
```suggestion
      # Insert vote_B and get related participant and proposal_url.
```

test/liquid_voting_web/absinthe/mutations/create_delegation/after_creating_related_votes_test.exs
```diff
@@ -0,0 +1,183 @@
...
+          }
+        }
+      }
+      """
+
+      {:ok, %{data: %{"createDelegation" => delegation}}} =
+        Absinthe.run(query, Schema, context: %{organization_id: vote_A.organization_id})
+
+      assert delegation["proposalUrl"] == proposal_A_url
+      assert delegation["votingResult"]["inFavor"] == 2
```
oliverbarnes, line 134: I'd add a comment with an explanation about the updated count here as well

lib/liquid_voting_web/resolvers/delegations.ex
oliverbarnes: I like the approach you went with, it's good to return the result with 0 votes.

lib/liquid_voting/voting.ex
oliverbarnes: 👍

oliverbarnes (changes requested):
This is excellent! 🦾  

Just have a few small suggestions about comments, otherwise it's good to ship

jinjagit:
Implemented all of @oliverbarnes'  suggestions.

oliverbarnes (approved):

PR #152, opened 2020-10-19
Update results on global delegation deletion

### LINKED ISSUE #44: Update all results that may be affected by deleting a global delegation?
Deletion of delegations is implemented by #43. 

Deleting a delegation for a specific proposal updates the voting result for that proposal.

Current deletion of global delegation doesn't update any results, as the ramifications have to be thought out carefully. There could be many results affected, and finding them could be a resource intensive procedure.

jinjagit:
Closes #44

### DIFF
--- a/lib/liquid_voting/voting_results.ex
+++ b/lib/liquid_voting/voting_results.ex
@@ -51,7 +51,7 @@ defmodule LiquidVoting.VotingResults do
   @doc """
   Publishes voting result changes to Absinthe's pubsub, so clients can receive updates in real-time
 
-  ## Examples
+  ## Example
 
       iex> publish_voting_result_change("https://www.medium/user/eloquent_proposal", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
       :ok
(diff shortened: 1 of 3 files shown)


lib/liquid_voting/voting_results.ex
```diff
@@ -67,6 +67,27 @@ defmodule LiquidVoting.VotingResults do
...
+  the absinthe layer lib/liquid_voting/reolvers/delegations.ex file. This ensures
+  that related voting results AND Absinthe's pubsub are updated accordingly.
+
+  ## Example
+
+      iex> publish_voting_result_changes_for_participant("377ead47-05f1-46b5-a676-f13b619623a7", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      :ok
+
+  """
+  def publish_voting_result_changes_for_participant(participant_id, organization_id) do
```
jinjagit, line 84: I chose to abstract out this def (which is called by both the delegations.ex resolver create_delegation & delete_delegation functions) and place it here, in LiquidVoting.VotingResults, as this seemed most logical to me. It could also live in the resolver file, however.  
Thoughts on this choice?

oliverbarnes: This falls in that gray area of concerns that Niklas brought up a while back: it is a `VotingResults` concern, semantically, but it's touching the absinthe layer from the business logic / data layer. And granted, we were already doing that with `VotingResults.publish_voting_result_change` beforehand.

We can go with this for now, but we do need to refactor soonish. This mixing of the layers is confusing

oliverbarnes (approved):
Looks good to me! I'll see if we still have Niklas' refactor issue lying around, or I'll open a new one.

lib/liquid_voting/voting_results.ex
jinjagit: > We can go with this for now, but we do need to refactor soonish. This mixing of the layers is confusing

Understood. We could look at this when we pair on #153 :+1:

PR #154, opened 2020-10-20
Simplify absinthe delegation deletion test

oliverbarnes:
Just some test cleanup I noticed could be done after merging https://github.com/liquidvotingio/api/pull/152

Removes extra request in one test and simplifies payload on setup request for another one.


jinjagit (approved):
LGTM

Issue #123, opened 2020-09-07
Decide on format for error handling


niklaslong:
As noticed in #120, we currently have different _shapes_  for the errors in the application: 
- `{:error, message, detail}`
- `{:error, %{message: message, detail: detail}`

This issue is meant to start the discussion around how we structure errors, abstractions on error handling and so on...

jinjagit:
I based error messages on the format I found in `assert details == %{email: ["is invalid"]}` from `test/liquid_voting_web/absinthe/mutations/create_participant_test.exs`  
I'm not sure, but that error looks like it is coming form an ecto email validation, via a changeset, which confuses me, if it is not in the format we desire.

jinjagit:
Useful / relevant posts & articles:  

1. https://elixirforum.com/t/best-practises-for-error-handling-elixir/2532/5 - especially [this post about error pattern matching](https://elixirforum.com/t/best-practises-for-error-handling-elixir/2532/6?u=jinjagit)

niklaslong:
Random thought: might be worth creating an `Error` struct(s)? I'm mentioning it here as it's actually something I've struggled with in other projects in the past as well, would be curious to hear thoughts on this. E.g.: 

```
{:error, %Error{message: "ran out of chocolate", detail: "there's never enough chocolate", ...}}
```

jinjagit:
So, here's my current idea:  

- Create an error handling module. All functions that expect an `{:ok, something}` also match for `{:error, some_error_stuff}` which is then passed on to this error handling module by bubbling the error up (as just `{:error, some_error_stuff}` to the highest level involved in the call.
- This error handling module can then match against all needed error formats. Typically, this would be `message: "message_string", details: "details_string"` (or an error struct, as suggested by Niklas).
- If, for example, the format `details: %{email: ["is invalid"]}` is returned by the phoenix email validation process, then we can add a match for this format in the error handling module, and then decode this into our standard format (e.g. `details: "email is invalid"`)
- This way, we only have to add any 'new / different' error formats to one place, all functions that finally return `:ok` or `:error` just match `{:error, some_error_stuff} -> error_handler(some_error_stuff).
- Niklas's error struct idea seems like a good one, and could be incorporated into this approach.
- the final error handling module pattern match could be a catch-all that simply turns whatever the (unexpectedly weird) error format is into a string, so that at least we will see the outputs as a return value in absinthe and hopefully think "We need to change that error format / explicitly parse it in the error handling module"

[...]

PR #162, opened 2020-11-03
Vote prevents conflicting delegation

### LINKED ISSUE #160: Proposal delegation creation should be prevented when vote for same participant (delegator) & proposal exists
Currently, the following sequence of actions fails, when it should succeed OR be halted (by an error - cannot create delegation due to vote for same proposal + delegator existing) at step 2.:  
```
Note: All actions in this example involve the same specific proposal.

1. createVote  
2. createDelegation (delegator = participant who just voted)
3. deleteDelegation (delete delegation just created)
4. createVote (try to create same vote as initial createVote)
=> "Could not create vote / Participant id has already been taken"
```
  
Not yet manually tested in case where no vote created first (but I suspect this does not fail and may already be covered by tests), nor manually tested with global delegation case.

jinjagit:
Closes #160

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -185,6 +185,8 @@ defmodule LiquidVoting.Delegations do
   def upsert_delegation(%{delegator_id: delegator_id, delegate_id: delegate_id} = attrs) do
     proposal_url = Map.get(attrs, :proposal_url)
 
+    if proposal_url != nil do check_vote_conflict(delegator_id, proposal_url) end
+
     Delegation
     |> where(delegator_id: ^delegator_id)
     |> Repo.all()
@@ -199,6 +201,13 @@ defmodule LiquidVoting.Delegations do
       {:error, %{message: message, details: details}} ->
         {:error, %{message: message, details: details}}
     end
+  end
+
+  defp check_vote_conflict(delegator_id, proposal_url) do
+    Voting.Vote
+    |> where(participant_id: ^delegator_id, proposal_url: ^proposal_url)
+    |> Repo.all()
+    |> IO.inspect()
   end
 
   # Resolves conflicting delegations (2 clauses).
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -223,7 +223,7 @@ defmodule LiquidVoting.Delegations do
         {
           :error,
           message: "Could not create delegation.",
-          details: "Vote for same delegator & proposal exists."
+          details: "Delegator has already voted on this proposal."
         }
 
       # Happy path: no conflicting vote found.
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -206,27 +206,29 @@ defmodule LiquidVoting.Delegations do
 
   # Checks for a conflicting vote, in the case where a proposal-specific delegation creation is attempted.
   #
-  # Returns {:ok} if delegation creation is for a global delegation, or if no conflicting vote is found.
+  # Returns {:ok} if delegation creation is for a global delegation.
+  #
+  # Or returns {:ok} if delegation creation is for a proposal delegation & no conflicting vote is found.
   # Returns an error, if a conflicting vote is found.
-  defp check_vote_conflict(delegator_id, proposal_url) do
-    if proposal_url != nil do
-      Vote
-      |> where(participant_id: ^delegator_id, proposal_url: ^proposal_url)
-      |> Repo.one()
-      |> case do
-        %Vote{} ->
-          {
-            :error,
-            message: "Could not create delegation.",
-            details: "Vote for same delegator & proposal exists."
-          }
+  defp check_vote_conflict(_delegator_id, _proposal_url = nil) do
+    {:ok}
+  end
 
-        # Happy path: no conflicting vote found.
-        nil ->
-          {:ok}
-      end
-    else
-      {:ok}
+  defp check_vote_conflict(delegator_id, proposal_url) do
+    Vote
+    |> where(participant_id: ^delegator_id, proposal_url: ^proposal_url)
+    |> Repo.one()
+    |> case do
+      %Vote{} ->
+        {
+          :error,
+          message: "Could not create delegation.",
+          details: "Vote for same delegator & proposal exists."
+        }
+
+      # Happy path: no conflicting vote found.
+      nil ->
+        {:ok}
     end
   end
 
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -215,10 +215,7 @@ defmodule LiquidVoting.Delegations do
   end
 
   defp check_vote_conflict(delegator_id, proposal_url) do
-    Vote
-    |> where(participant_id: ^delegator_id, proposal_url: ^proposal_url)
-    |> Repo.one()
-    |> case do
+    case Voting.get_vote_by_participant_id(delegator_id, proposal_url) do
       %Vote{} ->
         {
           :error,
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -7,6 +7,7 @@ defmodule LiquidVoting.Delegations do
 
   alias __MODULE__.Delegation
   alias LiquidVoting.{Repo, Voting}
+  alias Voting.Vote
   alias Ecto.Multi
 
   @doc """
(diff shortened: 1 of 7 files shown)


lib/liquid_voting/delegations.ex
```diff
@@ -185,19 +186,47 @@ defmodule LiquidVoting.Delegations do
...
+      end
+    end
+  end
+
+  # Checks for a conflicting vote, in the case where a proposal-specific delegation creation is attempted.
+  #
+  # Returns {:ok} if delegation creation is for a global delegation, or if no conflicting vote is found.
+  # Returns an error, if a conflicting vote is found.
+  defp check_vote_conflict(delegator_id, proposal_url) do
+    if proposal_url != nil do
```
jinjagit, line 212: Although I managed to avoid nested case statements in `upsert_delegation/1`, I used an `if` statement, which contains a nested case statement, in `check_vote_conflict/2`. This seems inelegant and unnecessarily opaque. Any ideas to improve this approach would be gratefully received.

lib/liquid_voting_web/resolvers/delegations.ex
```diff
@@ -66,6 +66,9 @@ defmodule LiquidVotingWeb.Resolvers.Delegations do
 
       {:ok, delegation}
     else
+      {:error, message: message, details: details} ->
```
jinjagit, line 69: I used the most basic error form for the error message `{:error, message: message, details: details}`, since this is the format we want all error messages to take. Unfortunately this means that, for now, I have added another error match in `resolvers/delegations.ex/create_delegation/3` to pass this error onward. These error pattern matches should be greatly reduced when #123 is addressed.

test/liquid_voting_web/absinthe/mutations/delete_delegation_test.exs
```diff
@@ -73,7 +73,7 @@ defmodule LiquidVotingWeb.Absinthe.Mutations.DeleteDelegationTest do
     test "updates a related voting result" do
       proposal_delegation = insert(:delegation_for_proposal)
 
-      vote =
+      _vote =
```
jinjagit, line 76: This underscore avoids a warning re. unused var, when running test(s).

lib/liquid_voting/delegations.ex
```diff
@@ -185,19 +186,49 @@ defmodule LiquidVoting.Delegations do
...
+  # Or returns {:ok} if delegation creation is for a proposal delegation & no conflicting vote is found.
+  # Returns an error, if a conflicting vote is found.
+  defp check_vote_conflict(_delegator_id, _proposal_url = nil) do
+    {:ok}
+  end
+
+  defp check_vote_conflict(delegator_id, proposal_url) do
+    Vote
+    |> where(participant_id: ^delegator_id, proposal_url: ^proposal_url)
+    |> Repo.one()
```
oliverbarnes, line 220: For better separation of concerns, I'd move this query to a function under `Voting`, say `Voting.get_vote_by_participant_id`

lib/liquid_voting_web/resolvers/delegations.ex
oliverbarnes: true, this is getting out of hand 😅 but good for now

oliverbarnes (changes requested):
Liking where this is going - just a few comments

jinjagit:
All suggested changes implemented.  
  
Note: I have just noticed that I have not been passing on the `organization_id` (in args) from `delegations.ex/create_delegation/1` -> `upsert_delegation/1` -> `check_vote_conflict/2` -> `get_vote_by_participant_id/2`, and thus all the db queries that those subsequent functions use will also search through entries for organizations other than the one in play at the time. Should we create an issue to add in and use the `organization_id` throughout this chain?

lib/liquid_voting/voting.ex
```diff
@@ -143,6 +143,24 @@ defmodule LiquidVoting.Voting do
...
+
+  ## Examples
+
+      iex> get_vote_by_participant_id("a6158b19-6bf6-4457-9d13-ef8b141611b4", "https://proposals.net/2")
+      => %Vote{}
+
+      iex> get_vote_by_participant_id("a6158b19-6bf6-4457-9d13-ef8b141611b4", "https://proposals.com/non-existant-proposal")
+      => nil
+  """
+  def get_vote_by_participant_id(participant_id, proposal_url) do
```
oliverbarnes, line 157: Should also be constrained by organization_id

oliverbarnes (changes requested):

oliverbarnes:
> **Note:** I have just noticed that I have not been passing on the `organization_id` (in args) from `delegations.ex/create_delegation/1` -> `upsert_delegation/1` -> `check_vote_conflict/2` -> `get_vote_by_participant_id/2`, and thus all the db queries that those subsequent functions use will also search through entries for organizations other than the one in play at the time. Should we create an issue to add in and use the `organization_id` throughout this chain?

Good eye. Yeah, if you could do that and submit a PR fixing it after this one, that'd be great

lib/liquid_voting/delegations.ex
```diff
@@ -7,6 +7,7 @@ defmodule LiquidVoting.Delegations do
 
   alias __MODULE__.Delegation
   alias LiquidVoting.{Repo, Voting}
+  alias Voting.Vote
```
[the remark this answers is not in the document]
jinjagit: It' still used at line 219...  
```
defp check_vote_conflict(delegator_id, proposal_url) do
    case Voting.get_vote_by_participant_id(delegator_id, proposal_url) do
      %Vote{} ->
        {
          :error,
          message: "Could not create delegation.",
          details: "Delegator has already voted on this proposal."
        }

      # Happy path: no conflicting vote found.
      nil ->
        {:ok}
    end
  end
```

jinjagit:
@oliverbarnes You said: "Is the var assignment needed at all?" 
  
I don't know why, but I can't quote or reply to that comment.  
 
I will remove the unused var name.

lib/liquid_voting/delegations.ex
```diff
@@ -185,19 +186,46 @@ defmodule LiquidVoting.Delegations do
...
-    |> case do
-      {:ok, delegations} ->
-        delegations
-        |> find_similar_delegation_or_return_new_struct(proposal_url)
-        |> Delegation.changeset(attrs)
-        |> Repo.insert_or_update()
-
-      {:error, %{message: message, details: details}} ->
-        {:error, %{message: message, details: details}}
+    with {:ok} <- check_vote_conflict(delegator_id, proposal_url) do
```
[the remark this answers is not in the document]
jinjagit: [`with`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1) doesn't require an explicit else block for errors, unless you want to specify a particular error.  

That's why I chose to use `this` - it enables the passing on of an error from check_vote_conflict/2, without creating a nested conditional structure.

The error is demonstrated in the 2 tests in this PR adds which assert this error is thrown, and also in manual testing.

lib/liquid_voting/voting.ex
jinjagit: I will do this immediately this PR is merged (see issue #164).

jinjagit:
"Good eye. Yeah, if you could do that and submit a PR fixing it after this one, that'd be great"  
  
Once again, I could not reply directly to this comment.  

I will do this immediately this PR is merged (see issue #164).

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
oliverbarnes: I know `with` doesn't require an `else` block :) But I'd still like to understand where the error tuple is handled. It should be explicitly visible

Btw, "thrown" is said of exceptions, not returns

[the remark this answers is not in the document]
jinjagit: You want an explicit else block in `upsert_delegation/1`?  Like this?
```
with {:ok} <- check_vote_conflict(delegator_id, proposal_url) do
      Delegation
      ...
      |> case do
        {:ok, delegations} ->
          ...

        {:error, %{message: message, details: details}} ->
          {:error, %{message: message, details: details}}
      end
else 
  {:error, message: message, details: details} ->
        {:error, message: message, details: details}
end
``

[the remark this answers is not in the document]
Comment (oliverbarnes) [reply lib/liquid_voting/delegations.ex:189]: No, I was just trying to understand code execution flow. It wasn't immediately obvious (to me) where the error went. Might just a be another symptom the error handling needs a refactor.

I've now fetched the PR branch locally and I'm having a look in my editor. What I'm understanding is `upsert_delegation/1` will return the error on line 141:

```elixir
|> Multi.run(:upsert_delegation, fn _repo, changes ->
  delegation_attrs
  ...
  |> upsert_delegation()
```

And get matched in line 148:

```elixir
{:error, :upsert_delegation, value, _} ->
  {:error, value}
```

I'm going to close this thread now and merge the PR, we can continue the code execution visibility discussion on refactor and error format issues.

oliverbarnes (approved):

lib/liquid_voting/delegations.ex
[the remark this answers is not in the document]
jinjagit: Yes, I think it will be much simpler when I refactor all the different error formats to one common one, as then we can simply have 2, chained, `with` statements here, with a single `else` statement for the common error format. :+1:

PR #167, opened 2020-11-08
Add organization_id to all child functions of upsert_delegations/3 that query db

### LINKED ISSUE #164: Pass on and use organization_id to filter db queries in children functions of create_delegation/1
`organization_id` (in args) isn't passed from `delegations.ex/create_delegation/1` -> `upsert_delegation/1` -> `check_vote_conflict/2` -> `get_vote_by_participant_id/2`, and thus all the db queries that those subsequent functions use will also search through entries for organizations other than the one in play at the time. 

These should all use `organization_id` to filter queries.

jinjagit:
Closes #164

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -183,18 +183,21 @@ defmodule LiquidVoting.Delegations do
       iex> upsert_delegation(%{field: bad_value})
       {:error, %Ecto.Changeset{}}
   """
-  def upsert_delegation(%{delegator_id: delegator_id, delegate_id: delegate_id} = attrs) do
+  def upsert_delegation(
+        %{delegator_id: delegator_id, delegate_id: delegate_id, organization_id: organization_id} =
+          attrs
+      ) do
     proposal_url = Map.get(attrs, :proposal_url)
 
-    with {:ok} <- check_vote_conflict(delegator_id, proposal_url) do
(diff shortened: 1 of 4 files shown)


oliverbarnes (changes requested):
LGTM, I'd just add a test that detects the bug this PR fixes

jinjagit:
> LGTM, I'd just add a test that detects the bug this PR fixes

Wouldn't it need to be multiple tests (since there are multiple functions involved)? And does that also imply we should add tests of all functions to check that they include filtering by organization_id?

oliverbarnes:
> Wouldn't it need to be multiple tests

No, just something as high-level as it makes sense, replicating the potential problem of delegations being affected across organizations. Just one test is enough to document the bug existed, and catching the problem in the future if a regression happens for any reason

jinjagit:
> > Wouldn't it need to be multiple tests
> 
> No, just something as high-level as it makes sense, replicating the potential problem of delegations being affected across organizations. Just one test is enough to document the bug existed, and catching the problem in the future if a regression happens for any reason

So, for example, a test that checks that deleting a delegation does not delete a similar delegation with a different organization_id (as it would if the organization_id was not used as a filter)?

oliverbarnes:
exactly 🎯

jinjagit:
Added `test "upsert_delegation/1 with proposal delegation does not conflict with vote with different organization_id"`

oliverbarnes (approved):

PR #182, opened 2021-01-20
Use AGPL 3 license

### LINKED ISSUE #176: Change license from MIT to AGPL
I've been reading up on the differences between them, and seems like MIT allows for a scenario where the codebase can be used commercially without contributing back any improvements made to it:

> One famous case is Apple taking BSD (which was licensed with a permissive license) and creating OSX based on it. They were not able to do that with Linux as Linux was not licensed with an MIT-like license.

> MIT license is great if you only care about the spread of your code, and not so great if you’d like to encourage more companies to contribute back to the world of open source. But there’s a license for that too.

AGPL prevents that:

> AGPL is designed to ensure corporations contribute back to the open source community even when running the software as a service in the cloud.

> If you used AGPL-licensed code in your web service in the cloud, you are required to open source it. It basically prevents corporations that never had any intention to contribute to open source from profiting from the open source work.

> It explicitly prohibits corporations from parasitically competing with an open source project. They won’t be able to take the code, make changes to it and sell it as a competing product without contributing those changes back to the original project.

https://plausible.io/blog/open-source-licenses

Aside from Linux, Wordpress also uses it.

Thoughts? Any unforeseen downsides to switching?

oliverbarnes:
Closes #176

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,19 +1,661 @@
-Copyright (c) 2019 Oliver Azevedo Barnes
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


davefrey (approved):
Merging with same comments as https://github.com/liquidvotingio/auth/pull/74

PR #202, opened 2021-03-10
Add opentelemetry_honeycomb.

jinjagit:
Followed guide at https://github.com/garthk/opentelemetry_honeycomb, including verification steps (with success).

Next steps?

- [ ] Get absinthe specific telemetry to opentelemetry
- [ ] Create dataset on Honeycomb from opentelemetry data


oliverbarnes:
Nice progress 🚀 

We'll also need to add Honeycomb's api key to our api deployment. Created an issue for it: https://github.com/liquidvotingio/deployment/issues/63

Let's just tackle absinthe telemetry in a separate issue and PR please, so we do things more incrementally and get a quicker win.

jinjagit:
Added [OtelGettingStarted.hello example](https://opentelemetry.io/docs/collector/getting-started/#getting-started).

This provides an automatic generation of some dummy data to export,
and currently exports to stdout using `:otel_exporter_stdout`.

Hopefully, this can be used to create a test dataset on Honeycomb
by switching to the Honeycomb exporter in `config.exs`, following
setting up the `HONEYCOMB_WRITEKEY` env var.

It might also provide a useful template for handling absinthe telemetry in preparation for export via OpenTelemetry.

This uses a release, defined in `def project` in `mix.exs`. Whilst this works, I am not at all familiar with using releases in Elixir, so I am a little concerned this approach might introduce some unwanted side-effects.

OtelGettingStarted_hello

oliverbarnes:
I was hoping this lib would already give us phoenix's telemetry automatically. Phoenix, Plug and Ecto all have telemetry built-in: https://thoughtbot.com/blog/instrumenting-your-phoenix-application-using-telemetry

jinjagit:
> I was hoping this lib would already give us phoenix's telemetry automatically. Phoenix, Plug and Ecto all have telemetry built-in: https://thoughtbot.com/blog/instrumenting-your-phoenix-application-using-telemetry

That would be great, and may well be true.

jinjagit:
I think this might be at the point where dummy data created in `lib/otel_getting_started.ex` will get sent by the OpenTelemetry export to Honeycomb. I am not sure if this will create a new dataset (with what name?) or I need to do some more configuring somewhere to associate this telemetry data with a specific (new or existing) dataset.

oliverbarnes (approved):

PR #206, opened 2021-03-12
Uncomment config :opentelemetry, :resource, settings.

jinjagit:
Try using these config settings.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -43,8 +43,8 @@ config :opentelemetry,
 # You can also supply opentelemetry resources using environment variables, eg.:
 # OTEL_RESOURCE_ATTRIBUTES=service.name=name,service.namespace=namespace
 
-# config :opentelemetry, :resource,
-#   service: [
-#     name: "service-name",
-#     namespace: "service-namespace"
-# ]
+config :opentelemetry, :resource,
+  service: [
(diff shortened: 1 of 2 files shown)


config/config.exs
```diff
@@ -43,8 +43,8 @@ config :opentelemetry,
...
 # OTEL_RESOURCE_ATTRIBUTES=service.name=name,service.namespace=namespace
 
-# config :opentelemetry, :resource,
-#   service: [
-#     name: "service-name",
-#     namespace: "service-namespace"
-# ]
+config :opentelemetry, :resource,
+  service: [
+    name: "service-name",
```
oliverbarnes, line 48: would be good to now what `resource`, `service` and `namespace` all refer to here. `service-name` and `service-namespace` look like placeholder values

oliverbarnes (changes requested):

config/config.exs
jinjagit: Agree. I have done some searching on that, but to no avail as yet.

jinjagit: Let's try without these (commented out again), but with dummy data?

oliverbarnes: worth a try 👍

oliverbarnes (approved):

PR #219, opened 2021-03-19
Include success or failure paths in traces.

jinjagit:
Includes trace attribute field of 'result' for displaying details of `:ok` or `:error` result, where applicable.

Includes all functions involved in successful create vote path, initiated via GraphQl `createVote` mutation, and specific failure paths (but not all, yet).

### DIFF
--- a/lib/liquid_voting/voting.ex
+++ b/lib/liquid_voting/voting.ex
@@ -96,10 +96,17 @@ defmodule LiquidVoting.Voting do
 
   """
   def list_votes_by_proposal(proposal_url, organization_id) do
-    Vote
-    |> where(proposal_url: ^proposal_url, organization_id: ^organization_id)
-    |> Repo.all()
-    |> Repo.preload([:participant])
+    Tracer.with_span "#{__MODULE__} #{inspect(__ENV__.function)}" do
+      Tracer.set_attributes([
+        {:request_id, Logger.metadata()[:request_id]},
+        {:params, [{:organization_id, organization_id}, {:proposal_url, proposal_url}]}
+      ])
+
+      Vote
+      |> where(proposal_url: ^proposal_url, organization_id: ^organization_id)
+      |> Repo.all()
+      |> Repo.preload([:participant])
+    end
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


jinjagit:
Example happy path trace (with some fields added to waterfall graph):  
  
Screenshot from 2021-03-19 20-34-09

Example sad path trace (with some fields added to waterfall graph):  

Screenshot from 2021-03-19 20-36-22

jinjagit:
Question:

Currently, I have chosen to report ":ok, \<name of action called that led to result\>" in a 'result' field.
For example `:ok, Voting.create_vote`
And for errors to report: ":error, \<name of action called that led to result\>, error message, error details" in the 'result' field.
For example:
```
:error, Voting.upsert_participant
message: Could not create vote with given email', details: '%{email: ["is invalid"]}'
```

I think this is more useful that reporting the actual value that accompanies `:ok`, since this often includes many 'not_loaded' statements and other ephemera - as well as the parameters, that are/were passed onto functions in the related traces (that will be grouped together in the waterfall view) and are given in the parameter field of those functions.

For example, upsert_participant gives an success tuple of:
```
{:ok, "%LiquidVoting.Voting.Participant{__meta__: #Ecto.Schema.Metadata<:loaded, \"participants\">, delegations_received: #Ecto.Association.NotLoaded<association :delegations_received is not loaded>, email: \"jj@somedomain.com\", id: \"d21feb63-0d1a-4059-a769-3b1663942dad\", inserted_at: ~N[2021-03-19 18:46:25], name: nil, organization_id: \"b7a9cae5-6e3a-48b1-8730-8b5c8d6c9b5a\", updated_at: ~N[2021-03-19 18:46:25], votes: #Ecto.Association.NotLoaded<association :votes is not loaded>}"}
```
...which does not seem very useful to record, whereas recording the name of the action that led to the result seems useful + more concise / readable.

Thoughts?

oliverbarnes:
I think this is fine, the approach is making sense to me and the resulting trace is looking pretty useful 👏  

I'm good to merge this, unless there's anything else you wanted to add to the PR

jinjagit:
I'm fine with  merging. I feel like I should be able to find a better method than `"... #{inspect(__ENV__.function)}"` and the other case where I have used `inspect` to create strings containing things that don't convert to strings using the `#{...}` approach, but I can do this in a separate PR (probably today).

oliverbarnes:
Yeah, these are improvements that can be done separately 👍  merging

oliverbarnes (approved):
Great work! 💯

PR #221, opened 2021-03-22
Add ecto telemetry using geometrics dependency.

jinjagit:
**TLDR:**  

Not ideal, as [geometrics](https://github.com/geometerio/geometrics) adds front-end tracing functionality which we don't need, and also enables phoenix telemetry which we already have. On the other hand, it [allegedly also ensures spans aren't orphaned](https://github.com/geometerio/geometrics/blob/main/guides/phoenix.md) "...if a process crashes or exits with an open span." This apparently would not occur if we simply added `opentelemetry_ecto`.

I tried adding [opentelemetry_ecto](https://github.com/opentelemetry-beam/opentelemetry_ecto) without the geometrics library (which includes `opentelemetry_ecto`), but to no avail - no Ecto trace spans nor fields were created.

Example of create vote happy path trace, including ecto queries:  
ecto_trace


jinjagit:
**DETAILS:**

Three dependencies specified in `mix.exs` required overriding dependency versions used geometrics:
The first 2, `:opentelemetry, "~> 0.5.0", override: true` & `:opentelemetry_api, "~> 0.5.0", override: true` are also needed when adding [opentelemetry_ecto](https://github.com/opentelemetry-beam/opentelemetry_ecto) without geometrics (which includes opentelemetry_ecto), as opentelemetry_ecto requires earlier versions (or beyond) of these 2 deps.
The third, `:opentelemetry_phoenix, "~> 0.2.0", override: true` is needed because it diverges from the geometrics phoenix dep (which specifies a path), resulting in this message on `mix deps.get` (until `override: true` is added):
```
Dependencies have diverged:
* opentelemetry_phoenix (Hex package)
  the dependency opentelemetry_phoenix in mix.exs is overriding a child dependency:

  > In mix.exs:
    {:opentelemetry_phoenix, "~> 0.2.0", [env: :prod, repo: "hexpm", hex: "opentelemetry_phoenix"]}

  > In deps/geometrics/mix.exs:
    {:opentelemetry_phoenix, [env: :prod, git: "https://github.com/opentelemetry-beam/opentelemetry_phoenix.git"]}

  Ensure they match or specify one of the above in your deps and set "override: true"
```

I could, perhaps, reverse engineer how geometrics gets opentelemetry_ecto to produce traces, whereas following the (very sparse) [guide in the repo readme](https://github.com/opentelemetry-beam/opentelemetry_ecto/blob/master/README.md) does not. I have tried various fairly simple variations and extensions of the guide, including some ideas taken from the geometrics code, all with no significant result. Having spent a couple of days on this, and given that the geometrics approach alleges to give us more robust span completion (in the case of a process crash or exit), maybe we should just go with this for now?

Also, opentelemetry for Elixir seems, in general, to be lacking in maturity, with many deps not yet at v. 1+, so all of this may get easier in the nearish future.

jinjagit:
**Git path in geometrics dependency:**

I needed to edit all occurences of `image: elixir:1.11.3-slim` to `image: elixir:1.11.3` to enable the dependencies to be built in the linting & tests actions in the `on_pull-request.yml`. I did the same for the `on_push.yml` but am unsure if more is needed to get the deps built in the `build-publish` job (do we need git added to the api docker image, for example?)

oliverbarnes (approved):

Issue #225, opened 2021-03-29
Add voting methods


oliverbarnes:
(This emerged on https://github.com/liquidvotingio/decidim-module-liquidvoting/issues/31#issuecomment-809493831)

Add a new `voting_method` graphql field to vote mutations and queries, and to voting results.

A string field mapping to entries in a new `voting_methods` db table (with matching Ecto schema) that gets upserted so we avoid the need for pre-seed or for an admin interface to manage it.

It would map to an association for `Vote` and `VoteResult` schemas. For `Vote`, for example:

```elixir
defmodule LiquidVoting.Voting.Vote do    
  schema "votes" do
    belongs_to :voting_methods, VotingMethods
```

Same for `VotingResult`.

~~> (@jinjagit) And this field would be optional? To allow our api to still work in it's purest, no voting method specified, way, as (for example) in the api readme examples. So we could use a create_vote api function, for example, that picks up the voting_method field if present and does the voting_method table stuff, and if no voting_method is specified the votes/results etc. are processed as they are now (without considering any voting_method).~~

~~👍  agreed, it should be optional. To avoid splitting db queries we can just have an empty entry in the db.~~

[EDIT: `voting_method` now required, not optional - see [comment below](https://github.com/liquidvotingio/api/issues/225#issuecomment-811911314)]

The new VotingMethod schema would also be constrained by organization id.

jinjagit:
Just to help me think this through, table would just have 2 columns:
```
voting_method   |   org_uuid  
----------------------------------------------------------------------
supports            |   02809347501934785
endorsements   |   02809347501934785
                          |   09708789079875676

jinjagit:
`voting_method` will be required when creating a vote.

In discussion with Oliver, we decided this was the easiest way to avoid the complexities of trying to create a unique index against possible null values (do-able, but then creates Postgrex exception, which would need special handling, if an empty string is given for voting_method.name) & to also avoid explicitly setting a default string value. This means a voting_method is required even when a voting context (e.g. proposal_url) only has one 'method'.

`voting_method` will be associated with a result, and will be required when initiating a result calculation (not yet implemented).

`voting_method` may be required for list_votes, though I am currently undecided on this, as I can see pros and cons in having both a list_all_votes (of all voting methods) and a list_votes_for_voting method action.

jinjagit:
Closing, as this was completed in #227

PR #227, opened 2021-03-31
Use voting_method for voting, delegations & results

jinjagit:
Aim: To pass `voting_method` via absinthe/GraphQl, and when it is included, record and use its value appropriately.

Addresses #225

### DIFF
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -53,10 +53,20 @@ defmodule LiquidVoting.Delegations do
 
   ## Examples
 
-      iex> get_delegation!("delegator@email.com", "delegate@email.com", "https://aproposal.com", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      iex> get_delegation!(
+        "delegator@email.com",
+        "delegate@email.com",
+        "method_A",
+        "https://aproposal.com",
+        "a6158b19-6bf6-4457-9d13-ef8b141611b4")
       %Delegation{}
 
-      iex> get_delegation!("participant-without-delegation@email.com", "some@body.com", "https://aproposal.com", "a6158b19-6bf6-4457-9d13-ef8b141611b4")
+      iex> get_delegation!(
+        "participant-without-delegation@email.com",
+        "some@body.com",
+        "method_A",
+        "https://aproposal.com",
+        "a6158b19-6bf6-4457-9d13-ef8b141611b4")
       ** (Ecto.NoResultsError)
 
   """
@@ -79,6 +89,7 @@ defmodule LiquidVoting.Delegations do
       proposal_url: proposal_url,
       organization_id: organization_id
     )
+    |> Repo.preload([:voting_method])
   end
 
   @doc """
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -182,7 +182,27 @@ defmodule LiquidVoting.Delegations do
   end
 
   def create_delegation(%{delegator_id: _, delegate_id: _} = attrs) do
-    upsert_delegation(attrs)
+    # TODO: Refactor common parts of 2 clauses of this function.
+    voting_method_id =
+      if Map.get(attrs, :proposal_url) do
+        {:ok, voting_method} =
+          VotingMethods.upsert_voting_method(%{
+            name: Map.get(attrs, :voting_method),
+            organization_id: attrs.organization_id
+          })
+
+        voting_method.id
+      else
+        nil
+      end
+
+    attrs = Map.put(attrs, :voting_method_id, voting_method_id)
+
+    {:ok, delegation} = upsert_delegation(attrs)
+
+    delegation = Repo.preload(delegation, [:voting_method], force: true)
+
+    {:ok, delegation}
   end
 
   @doc """
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -218,9 +218,6 @@ defmodule LiquidVoting.Delegations do
     proposal_url = Map.get(attrs, :proposal_url)
     voting_method_id = Map.get(attrs, :voting_method_id)
 
-    IO.puts("=============== upsert delegation attrs ================")
-    IO.inspect(attrs)
-
     with {:ok} <-
            check_vote_conflict(delegator_id, voting_method_id, proposal_url, organization_id) do
       Delegation
@@ -230,7 +227,11 @@ defmodule LiquidVoting.Delegations do
       |> case do
         {:ok, delegations} ->
           delegations
-          |> find_similar_delegation_or_return_new_struct(proposal_url, organization_id)
+          |> find_similar_delegation_or_return_new_struct(
+            proposal_url,
+            voting_method_id,
+            organization_id
+          )
           |> Delegation.changeset(attrs)
           |> Repo.insert_or_update()
 
@@ -314,13 +315,16 @@ defmodule LiquidVoting.Delegations do
   #  empty Delegation struct.
   #
   # Used by upsert_delegation/1 (above.)
-  defp find_similar_delegation_or_return_new_struct(delegations, proposal_url, organization_id) do
-    IO.puts("=============== find similar delegation attrs ================")
-    IO.puts("proposal_url: #{proposal_url}")
-
+  defp find_similar_delegation_or_return_new_struct(
+         delegations,
+         proposal_url,
+         voting_method_id,
+         organization_id
+       ) do
     delegations
     |> Enum.filter(fn d ->
-      d.proposal_url == proposal_url and d.organization_id == organization_id
+      d.proposal_url == proposal_url and d.voting_method_id == voting_method_id and
+        d.organization_id == organization_id
     end)
     |> case do
       # Delegation (of same type) not found, so we build one
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -204,7 +204,7 @@ defmodule LiquidVoting.Delegations do
       end
   end
 
-  # If a proposal_url is specified, upserts a voting_method and returns the voting_method_id,
+  # If a proposal_url is specified, upserts a voting_method and returns the voting_method_id.
   # If a proposal_url is NOT specified, simply returns voting_method_id == nil.
   defp upsert_voting_method_and_get_id(attrs) do
     voting_method_id =
--- a/lib/liquid_voting/delegations.ex
+++ b/lib/liquid_voting/delegations.ex
@@ -135,10 +135,20 @@ defmodule LiquidVoting.Delegations do
     delegate_attrs = %{email: args.delegate_email, organization_id: args.organization_id}
     delegation_attrs = Map.take(args, [:organization_id, :proposal_url])
 
-    # Also get voting_method (==name) from args and add to delegation attrs (could be nil)
-    # If proposal_url not nil
-    #   Get voting_method_name from args if not already set to nil when no string provided)
-    #   Upsert voting method (and get id from returned struct? - not sure we need this here)
+    # If a proposal_url is specified, we upsert a voting_method and return the voting_method_id,
+    # If a proposal_url is specified, we simply return voting_method_id == nil.
+    voting_method_id =
+      if Map.get(args, :proposal_url) do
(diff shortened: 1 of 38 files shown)


jinjagit:
Currently it is possible to create a null value on the db for voting_method.name by passing in an empty string "" in absinthe `createVote` mutation. Using `null: false` on the db field (via a migration) will prevent this, but will also throw a Postgrex error, which results in an exception (exit).

Any ideas on best practice to prevent this? Maybe at the changeset level? With a `validate_string` function, that returns a custom error?

jinjagit:
Summary of progress:  
- All tests now green, though I expect to add more specific tests as work continues / when work is nearing completion.
- Creation of vote now requires a `voting_method` parameter (at core and absinthe layers), which equates to `voting_method.name`.
- `votingMethod` (and its fields and values) now exposed as associated with any existing vote(s) at absinthe layer = can be queried in query of vote(s)
- results calculation does not yet involve `voting_method`
- `list_votes` does not yet make use of `voting_method`

@oliverbarnes @davefrey This feels like a good point to get some feedback / ask & answer questions. I could participate in a meeting about this, if preferred.
 
Unfortunately, I cannot submit for review as merging this will break our Decidim demo (and I can't really fix that until the api is ready to work fully in `voting_method` mode)

davefrey:
Hi Simon, congrats!

I can definitely review in the morning...I won't have the time to look carefully today.

You might describe more your approach and decisions in the PR description, that way I can understand your intent rather than inferring from the code: it will help me separate what you wanted to do from whether the code matches.

For example my first questions...again I can infer from the code, but tell us the story :-)
* is this voting method required or do we have a default on the api (does the client have to send one)
* is the voting result that we return for (almost) all mutators meant to include a count just for that voting method (just the supports count for example), or a comprehensive result for the proposal url (counts for endorsements and supports)? 

Apologies if this is all laid out in previous slack or git issue conversations, ideally I'd be current on all that. And if your implementation intention is set out in the issue that's 👍 for me...closely reading the issue is part of PR review, and I'll get there.

jinjagit:
> You might describe more your approach and decisions in the PR description, that way I can understand your intent rather than inferring from the code: it will help me separate what you wanted to do from whether the code matches.
> 
> For example my first questions...again I can infer from the code, but tell us the story :-)
> 
>     * is this voting method required or do we have a default on the api (does the client have to send one)

- Required when creating a vote. In discussion with Oliver, we decided this was the easiest way to avoid the complexities of trying to create a unique index against possible null values (do-able, but then creates Postgrex exception, which would need special handling, if an empty string is given for `voting_method.name`) & to also avoid explicitly setting a default string value. This means a `voting_method` is required even when a voting context (e.g. `proposal_url`) only has one 'method'.
- Is associated with a `result`, and will be required when initiating a result calculation (not yet implemented).
- May be required for `list_votes`, though I am currently undecided on this, as I can see pros and cons in having both a `list all votes (of all voting methods) _and_ a 'list votes by voting methods` action.

> 
>     * is the voting result that we return for (almost) all mutators meant to include a count just for that voting method (just the supports count for example), or a comprehensive result for the proposal url (counts for endorsements and supports)?
- Voting results will be specific to a `voting_method` (and `proposal_url`).

I think your questions and my answers here cover the main aspects of 'the story' so far. Feel free to ask more, though ;-)

jinjagit:
`voting_method` will be required when creating a vote.

In discussion with Oliver, we decided this was the easiest way to avoid the complexities of trying to create a unique index against possible null values (do-able, but then creates Postgrex exception, which would need special handling, if an empty string is given for voting_method.name) & to also avoid explicitly setting a default string value. This means a voting_method is required even when a voting context (e.g. proposal_url) only has one 'method'.

`voting_method` will be associated with a result, and will be required when initiating a result calculation (not yet implemented).

`voting_method` may be required for list_votes, though I am currently undecided on this, as I can see pros and cons in having both a list_all_votes (of all voting methods) and a list_votes_for_voting method action.

oliverbarnes:
This is looking good 🚀  I'll review again once the voting result part is done.

Good point about this only being mergeable when another PR on the module side consuming the new field is ready as well, we'll need to merge them both in tandem.

The only thing here that got me scratching my head so far is the updated index - I'm wondering if it's too granular. They have a cost in write time the bigger they get, and sometimes that outweighs the benefit. 

But I honestly need to refresh up on this topic, so I can't say for sure. This is good homework for you, @jinjagit: reading up on composite index best practices and give us a tldr on current state of our indexing, and then revisit the indexing change in this PR in light of what you find

jinjagit:
> @oliverbarnes updated index - I'm wondering if it's too granular. They have a cost in write time the bigger they get, and sometimes that outweighs the benefit.

Thanks for the input. I'll look into it now.

jinjagit:
> @Oliver reading up on composite index best practices and give us a tldr on current state of our indexing, and then revisit the indexing change in this PR in light of what you find

I have included a TLDR in a new issue: Review composite indexing #229.
On balance, I think this merits being a separate issue, avoiding too much scope creep in this PR.

Thoughts?

oliverbarnes:
> I have included a TLDR in a new issue: Review composite indexing #229.
> On balance, I think this merits being a separate issue, avoiding too much scope creep in this PR.

Agree 👍

jinjagit:
Note: When I have got `calculate_results`, etc. working with `voting_method` and passing tests, i will rename the `voting_method` param to `voting_method_name` internally, in core functions etc. to disambiguate from the `voting_method` struct. I will still use `votingmethod` as the param name for `voting_method.name` in absinthe queries & mutations, however, as this reads better at the absinthe layer (and when consumed by other services) which doesn't need to know about the underlying `voting_method` struct.

jinjagit:
Status update:

Basic `createVote`, `deleteVote` mutations + `votingResult` queries now working together as intended at absinthe layer, as can be seen in the following example 4 mutations (with queries of `votingResult` returned for each). Note the different 2 `votingMethod` names used and the related `votingResult` changes:
```
---- fred creates a vote for votingMethod "method_A" -------------------------------

mutation {
  createVote(votingMethod: "method_A", participantEmail: "fred@g.com", proposalUrl:"https://super-duper-somewhat-long-proposal-name", yes: true) {
    yes
      proposalUrl
      votingResult {
        inFavor
        against
      }
    }
}

=> 
{
  "data": {
    "createVote": {
      "proposalUrl": "https://super-duper-somewhat-long-proposal-name",
      "votingResult": {
        "against": 0,
        "inFavor": 1
      },
      "yes": true
    }
  }
}

---- fred creates a vote for votingMethod "method_B", same proposal ----------------

mutation {
  createVote(votingMethod: "method_B", participantEmail: "fred@g.com", proposalUrl:"https://super-duper-somewhat-long-proposal-name", yes: true) {
    yes
      proposalUrl
      votingResult {
        inFavor
        against
      }
    }
}

=>
{
  "data": {
    "createVote": {
      "proposalUrl": "https://super-duper-somewhat-long-proposal-name",
      "votingResult": {
        "against": 0,
        "inFavor": 1
      },
      "yes": true
    }
  }
}

---- ana creates a vote for votingMethod "method_A", same proposal -----------------

mutation {
  createVote(votingMethod: "method_A", participantEmail: "ana@g.com", proposalUrl:"https://super-duper-somewhat-long-proposal-name", yes: true) {
    yes
      proposalUrl
      votingResult {
        inFavor
        against
      }
    }
}

=>

{
  "data": {
    "createVote": {
      "proposalUrl": "https://super-duper-somewhat-long-proposal-name",
      "votingResult": {
        "against": 0,
        "inFavor": 2
      },
      "yes": true
    }
  }
}

---- fred deletes a vote for votingMethod "method_A", same proposal ----------------

mutation {
  deleteVote(votingMethod: "method_A", participantEmail: "fred@g.com", proposalUrl:"https://super-duper-somewhat-long-proposal-name") {
    yes
          proposalUrl
          votingResult {
            inFavor
            against
          }
        }
}

=>

{
  "data": {
    "deleteVote": {
      "proposalUrl": "https://super-duper-somewhat-long-proposal-name",
      "votingResult": {
        "against": 0,
        "inFavor": 1
      },
      "yes": true
    }
  }
}
```

Still lots of loose ends to tie up, but core new feature functionality is largely in place.

jinjagit:
Next step:  

Use `voting_method` when creating a specific (non-global) delegation. This avoids needing to return a list of results in relation to a `proposal_url`, as would be the case if one wanted to query result that could be changed by a delegates votes when making a proposal-specific delegation.

This will mean I should undo some recent changes, including adding a `list_results_for_proposal_url/3` function.

We need to decide at which level (core vs. absinthe) we wish to enforce the `proposal_url` + `voting_method` params, OR neither, when creating a delegation.

jinjagit:
We decided to include a 'default' value for `voting_method.name` when inserting a `voting_method`. Thus, when we create a delegation without a `proposal_url` we are still in 'global' land, but when we create a proposal-specific delegation, we use a `voting_method.name` to upsert `voting_method` (? we supply/insert 'default' name value if `voting_method.name` not in params? - this may clarify when implementing).

oliverbarnes:
> (? we supply/insert 'default' name value if `voting_method.name` not in params, or we enforce both params should be present: `proposal_url` && `voting_method.name`? - this may clarify when implementing).

We don't enforce voting_method presence in the graphql params - we just assume 'default' when not presence. DB will take of upserting the value when not present.

jinjagit:
> We don't enforce voting_method presence in the graphql params - we just assume 'default' when not presence. DB will take of upserting the value when not present.

OK. :+1: Also same logic with `create_vote`? Now we don't need to enforce `voting_method` in params when creating a vote?

oliverbarnes:
same 👍  that's a big reason to have the smart default, for use cases where there's only a single type of vote

jinjagit:
> This here is at the Absinthe resolver level :)
> 
> DB level would be this: https://hexdocs.pm/ecto_sql/Ecto.Migration.html#add/3-examples

I have set it at db level, and also just removed the absinthe resolver setting thing, as I checked the db does indeed still set the default value (_if_ we provide it with a nil value for the field)

jinjagit:
Note: Because we decided to use a default value for `votingMethod` `:name`, all the examples in the readme still work without any need to edit them :+1:

jinjagit:
[Video summary](https://www.youtube.com/watch?v=-mcbeFGFWvA) on YouTube
[Video with transcript](https://share.descript.com/view/ZWW2YlEheLp) on Descript

@oliverbarnes @davefrey Please review. Thanks!

DO NOT MERGE until changes to Decidim demo made to accommodate this new API functionality (or will break the demo)

jinjagit:
This should not break the demo if merged, since it will upsert `votingMethod: "default"` to our API database, for the current proposal support integration, without any need to edit the demo code and the graphQL mutations and queries it already uses.

`votingMethod: "supports"` only needs adding when we also use `votingMethod "endorsements"`.

jinjagit:
This should not break the demo if merged, since it will `upsert votingMethod: "default"` to our API database, for the current proposal support integration,and `votingResults` with no specified `votingMethod` should also return results for  `votingMethod: "default"`, without any need to edit the demo code and the graphQL mutations and queries it currently uses.

oliverbarnes (approved):
Thanks for all the work put into this, @jinjagit!

This is a big PR, but unavoidable due to the new voting_method being an almost global constraint, and much of the changes would break everything else if submitted separately.

Some of the api codebase was already due some refactoring and optimization, and some of the changes here increase that need a bit, but that tech debt is better tackled separately, in discrete and small PRs. They also take lower priority to moving ahead with integrating Decidim proposal endorsements, which depends on this PR.

I'm just going to note these small things here, for reference when tackling that:

- Repo preloads can take collections of associations, so multiple calls to it can be merged into one
- Some separate db queries might be more efficient if merged into a single one using joins
- Some function signatures are getting pretty big
- Some function are getting pretty complex, and this is already captured in some discussions here in the repo

And a subtle point: voting method sometimes comes as an argument in function signatures before proposal_url, when it's a constraint that most of the time will just use the default, and our domain it has a lower place - proposal, vote and delegation info pertain to the user actions (the user story), whereas voting method and organization id are more platform constraints - the user isn't choosing either in their voting and delegations, they're just given.

It's subtle but helps navigate the code if this hierarchy is clear and consistent :)

Going ahead and merging. Great work 👏

jinjagit:
Great feedback. Thanks! I'll create a/some issue(s) to cover the points you make.

PR #241, opened 2021-04-14
Bugfix: Preload voting_method when create vote for delegator.

jinjagit:
This fixes case where a voter is already the _delegator_ in a delegation for the same proposal (`proposal_url` + `voting_method`).

Prior to this fix, the process of searching for, and deleting the related delegation, in `create_vote/1` in `voting.ex`, returned a vote without the associated `voting_method` being preloaded, which would cause a `createVote` mutation to fail due to the associated `voting_method` being unavailable.

Adds a preload of `voting_method` when the vote is returned in this case.

Also adds a test of the case that highlighted this bug + a similar test of creating a vote for the _delegate_, following a proposal-delegation creation.

### DIFF
--- a/lib/liquid_voting/voting.ex
+++ b/lib/liquid_voting/voting.ex
@@ -50,8 +50,12 @@ defmodule LiquidVoting.Voting do
                    organization_id: attrs[:organization_id]
                  ) do
               case Delegations.delete_delegation(delegation) do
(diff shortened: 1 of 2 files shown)


oliverbarnes (approved):
Nice quick catch and fix 💯

